PR #7, opened 2018-03-13
Moved login information to config.py

AustinSanders:


bin/ChecksumUpdateProcess.py
```diff
@@ -19,6 +19,7 @@
 
 
 import pdb
+from config import *
```
jlaura: `*` style imports are generally considered poor form since another dev. really doesn't know what all is in the config namespace without going and looking at that file.  Perhaps alter the places where things in the config namespace are being hit to `config.<whatever>`

bin/ChecksumUpdateQueueing.py
```diff
@@ -19,6 +19,8 @@
 
 import pdb
 
+from config import *
```
jlaura: as above re: `*`

bin/ChecksumUpdateQueueing.py
```diff
@@ -74,7 +76,13 @@ def main():
 
 
     try:
-        engine = create_engine('postgresql://pdsdi:dataInt@dino.wr.usgs.gov:3309/pds_di_prd')
+        engine = create_engine('postgresql://{}:{}@{}:{}/{}'.format(pdsdi_user,
```
jlaura: How about we start a small refactor here.  Create a module with a `db.py file that has the logic for this connection.  It looks identical to the connection logic above.  Then you can centralize the connection logic in a single place.

bin/DIprocess.py
```diff
@@ -19,6 +19,7 @@
 
 
 import pdb
+from config import *
```
jlaura: Same

bin/DIprocess.py
```diff
@@ -72,7 +73,11 @@ def main():
     logger.info('Starting DI Process')
  
     try:
-        engine = create_engine('postgresql://pdsdi:dataInt@dino.wr.usgs.gov:3309/pds_di_prd')
+        engine = create_engine('postgresql://{}:{}@{}:{}/{}'.format(pdsdi_user,
+                                                                    pdsdi_pass,
+                                                                    pdsdi_host,
+                                                                    pdsdi_port,
+                                                                    pdsdi_db))
```
jlaura: Same refactor.

bin/DIqueueing.py
```diff
@@ -73,7 +75,12 @@ def main():
 
 
     try:
-        engine = create_engine('postgresql://pdsdi:dataInt@dino.wr.usgs.gov:3309/pds_di_prd')
+        engine = create_engine('postgresql://{}:{}@{}:{}/{}'.format(pdsdi_user,
+                                                                    pdsdi_pass,
+                                                                    pdsdi_host,
+                                                                    pdsdi_port,
+                                                                    pdsdi_db))
+
```
jlaura: I am not going to put any more of these in, you get the idea.  This will probably require a setup.py as well, since we are going to install the module into the namespace - I can assist with what this looks like.  Maybe even good to have everyone sit in a room for 30 while we all talk about this. @USGSsakins - what do you think?

bin/UPC_test.py
```diff
@@ -10,13 +10,20 @@
 from sqlalchemy.orm.util import *
```
jlaura: I know I bagged on the `*` imports above.  Since this is refactoring, I would not worry about touching `*` imports that we didn't introduce just yet.  Feel free to open a ticket though - this might be an ideal first(ish) code thing for @eabovre to look at.

bin/test_DBadd.py
```diff
@@ -37,7 +37,13 @@ def main():
     workarea = '/scratch/pds_services/workarea/'
 
     Base = declarative_base()
-    engine = create_engine('postgresql://upcmgr:un1pl@c0@dino.wr.usgs.gov:3309/upc_dev')
+    engine = create_engine('postgresql://{}:{}@{}:{}/{}'.format(upcdev_user,
+                                                                upcdev_pass,
```
jlaura: Perhaps this means that a different function should be in db that connects with these other config credentials or perhaps the credentials are stored in a dict by name (where value is the list of credentials) for easy access?

AustinSanders:
Do not merge -- I need to update calls to db_connect so that they handle the output properly.  Right now we'll end up with "too many values to unpack" or similar.

AustinSanders:
Do not merge until 20073e6 is tested

AustinSanders:
Sorry about multiple "initial models.py" - I thought the previous pushes were to a separate branch, but they ended up here.

@jlaura or @USGSsakins when you get a chance, can you take a look at the models.py to see if it looks like I'm on the right track?  The singledispatch stuff is a little hairy because the keytype parameter is specified as a string description (e.g. 'Boolean' or 'Integer') rather than an instance of a type.  I'm not sure if the workaround qualifies as "cool programming magic" or if it's just the non-canonical way of doing something that should be simple.

bin/models.py
```diff
@@ -0,0 +1,155 @@
+from sqlalchemy.ext.declarative import declarative_base, DeferredReflection
```
jlaura: We should not have a `models.py` inside of `bin`.

bin/models.py
```diff
@@ -0,0 +1,155 @@
...
+#  are prefixed with 'meta' (meta_precision, meta_integer, etc.),
+#  but it doesn't make much sense as is.
+def meta(keytype, *args, **kwargs):
+    """ Returns a table based on the keytype description.
+    
+    Parameters
+    ----------
+    keytype : str
+        A type description of the table's value.
+    *args
```
jlaura: No need to include `*args`/`**kwargs`, they are canonical.

bin/models.py
```diff
@@ -0,0 +1,155 @@
...
+    -------
+    out : :class:`.meta_string` or :class:`.meta_precision` or :class:`.meta_integer` or :class:`.meta_time` or :class:`.meta_boolean`
+        A SQLAlchemy table with type specific column specification.
+    """
+    try:
+        # Evaluate string type description to yield formal type
+        return _meta(eval(keytype.capitalize()), *args, **kwargs)
+    except NameError:
+        print('Keytype {} not found\n'.format(keytype))
+        return None
```
jlaura: This never returns because an error is raised.

bin/models.py
```diff
@@ -0,0 +1,155 @@
...
+    ----
+    This function differs from the :meth:`.meta` function in that it
+    takes a formal type as its first argument, while the non-underscored
+    version handles string type descriptions.
+    """
+    raise(NotImplementedError)
+
+@_meta.register(String)
+def _(c_type, **kwargs):
+    return meta_string(**kwargs)
```
jlaura: Class names need to be capitalized - this is really hard to understand what is going on.

bin/models.py
```diff
@@ -0,0 +1,155 @@
...
+    __tablename__ = 'meta_geometry'
+    upcid = Column(Integer, primary_key=True)
+    typeid = Column(Integer)
+    value = Column(Geometry('geometry'))
+
+if __name__ == "__main__":
+    t1 = meta('Boolean', upcid=1, typeid=0, value=True)
+    t2 = meta('string', upcid=2, typeid=1, value = 'StringExample')
+    t3 = meta('Integer', upcid=3, typeid=2, value = 1234)
+    t4 = meta('double', upcid=4, typeid=3, value = 5.6789)
```
jlaura: I am going to go dig into the source to see where these `meta_*` are being used. I am not quite sure what we are looking at here.  This looks like a hack to get introspective geometries - is that the case?  If so, this is way too much boiler plate / brittleness to accomplish that.

jlaura: I should also add - AFAIK, the canonical approach to do this is below.

`cls = type(str(tablehash), (basegeomcls, db.Model,), {'__tablename__':tablehash, '__table_args__' : {'extend_existing': True}})`

With the documentation for `type` coming from here: https://docs.python.org/3.3/library/functions.html#type

This is using `type` as a factory. The first arg is the class name, second are the base classes, and the third are the args for the class `__dict__`.

PR #11, opened 2018-04-05
DIqueueing NumPy Documentation

eabovre:
My first pull request! First part of DIqueueing documented.


bin/DIqueueing.py
```diff
@@ -25,18 +25,58 @@
...
+        of the namespace
+    """ 
     def __init__(self):
         pass
 
     def parse_args(self):
+        """
+        Defines how to parse single command-line argument
 
-        parser = argparse.ArgumentParser(description='PDS DI Data Integrity')
```
AustinSanders: We talked about the accidental removal of this line - make sure you add it back in future revisions.  Do not merge until this is addressed.

bin/DIqueueing.py
```diff
@@ -25,18 +25,58 @@
...
+
+    Attributes
+    ----------
+    archive : str
+    volume : str
+    jobarray : str
+
+    Methods
+    -------
+    _init_(self)
```
AustinSanders: The function should be documented as __init__ (surrounded by double underscores).

bin/DIqueueing.py
```diff
@@ -25,18 +25,58 @@
...
+        constructor
+    parse_args(self)
+        Converts argument strings into objects that are attributes
+        of the namespace
```
AustinSanders: I think that these are attributes of the object - not attributes of the namespace.

bin/DIqueueing.py
```diff
@@ -25,18 +25,58 @@
...
+    """ 
     def __init__(self):
         pass
 
     def parse_args(self):
+        """
+        Defines how to parse single command-line argument
 
-        parser = argparse.ArgumentParser(description='PDS DI Data Integrity')
+        Parameters
```
AustinSanders: The parse_args function does not take parameters.  The parameters section of documentation is intended to describe values that must be supplied to the function when it is called.

For example, if the function definition is `do_stuff(self, a, b, c): `   then you should document a, b, and c.  Since `parse_args(self)` does not require any information other than 'self,' there's nothing to document.

bin/DIqueueing.py
```diff
@@ -25,18 +25,58 @@
...
+        dest : str
+            name of attribute to be added to parser
+        
+        required : str
+            makes optional arguments required
 
+        help
+            contains brief description of argument
+
+        Returns
```
AustinSanders: The parse_args function does not have a return value.

bin/DIqueueing.py
```diff
@@ -25,18 +25,58 @@
 import pdb
 
 class Args:
+    """
+    Creates an object 'parser', which is populated by the attributes
```
AustinSanders: An 'Args' object is intended to store / provide access to command line arguments.  There is likely no need to discuss the 'parser' object at this level.

bin/DIqueueing.py
```diff
@@ -25,18 +25,58 @@
...
+    parse_args(self)
+        Converts argument strings into objects that are attributes
+        of the namespace
+    """ 
     def __init__(self):
         pass
 
     def parse_args(self):
+        """
+        Defines how to parse single command-line argument
```
AustinSanders: This function is responsible for parsing _all_ command-line arguments - not just a single argument.

jlaura:
Awesome to see this.  Very nice job!

bin/DIqueueing.py
eabovre: What would be attributes of the name space/are there any? Also, where would I document these attributes/should I?

bin/DIqueueing.py
eabovre: Okay, that makes sense. I guess I am getting confused because in this script none of the functions have arguments or they are just "self". The parameters I documented would be parameters for ArgumentParser() right?

PR #21, opened 2018-04-19
First step out of the skeleton phase

eabovre:
First go of adding types and small descriptions to attributes and parameters.

### DIFF
--- a/bin/BataMOWJobber.py
+++ b/bin/BataMOWJobber.py
@@ -21,12 +21,20 @@ import pdb
 
 class jobXML(object):
     """
+
     Attributes
     ----------
-    root
+    root:string
+        parses an XML section from a string constant
+        returns an Element instance
     """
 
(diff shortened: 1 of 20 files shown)


bin/BataMOWJobber.py
```diff
@@ -21,12 +21,20 @@
 
 class jobXML(object):
     """
+
     Attributes
     ----------
-    root
+    root:string
```
AustinSanders: Unless 'string' is of a different type than the standard 'str,' we should use 'str.'

jlaura:
@AustinSanders @USGSsakins Let's get this reviewed and merged?  Just of that the review is not epic.

@eabovre What do you think about this workflow. (1) Identify which file you want to work on, (2) open a new issue, (3) make the changes, (4) create a new PR, (5) update as necessary, (6) repeat?

PR #48, opened 2018-05-07
Finishing touches to HPCjob.py

eabovre:


eabovre:
I've been trying to make new branches for each commit so they are separated in the PR, but I haven't figured out how to do that correctly yet. Sorry about all of the changes smashed into one PR!

jlaura:
@USGSsakins If I were @eabovre I would be making a local branch for each change on my side, creating the PR from my local branch into this dev branch, and then removing my local branch when I rebased in the merged changes.  This is really clean and let's changes be isolated locally.

PR #53, opened 2018-05-11
environment.yml created, Loggy.py documentation finished

eabovre:
Loggy.py documentation got stuck on this branch without me realizing it. I finally created the environment.yml file. Let me know if something should be changed.

### DIFF
--- a/bin/Loggy.py
+++ b/bin/Loggy.py
@@ -8,6 +8,8 @@ from SubLoggy import *
 
 class Loggy(object):
     """
+    Object that has the attributes described below
+
     Attributes
     ---------
     Lfile : dict
--- a/bin/Loggy.py
+++ b/bin/Loggy.py
@@ -8,8 +8,8 @@ from SubLoggy import *
 
 class Loggy(object):
     """
-    Object that has the attributes described below
-
+    Writes to the logs file.
+    
     Attributes
     ---------
     Lfile : dict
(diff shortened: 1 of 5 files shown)


bin/Loggy.py
```diff
@@ -8,6 +8,8 @@
 
 class Loggy(object):
     """
+    Object that has the attributes described below
```
USGSsakins: The Loggy class is for writing to the logs file, add some kind of description to that effect.

environment.yml
```diff
@@ -0,0 +1,15 @@
+name: PDS-Pipelines
+dependencies:
+    - python=2
+    - pvl
+    - xml
```
AustinSanders: Remove the following modules:

- [ ] xml (built in)
- [ ] shutil (built in)
- [ ] config (local / shipped with package)
- [ ] squalchemy (duplicate of sqlalchemy with spelling error)
- [ ] pdb (built in)

environment.yml
```diff
@@ -0,0 +1,15 @@
...
+    - lxml
+    - shutil
+    - pysis
+    - pytz
+    - sqlalchemy
+    - config
+    - redis
+    - squalchemy
+    - geoalchemy2
+    - pdb
```
AustinSanders: should also include pip + a command to install pysis using pip instead of conda.

PR #63, opened 2018-05-30
Removed unnecessary dependencies and added a pip section

eabovre:


environment.yml
```diff
@@ -1,15 +1,12 @@
 name: PDS-Pipelines
 dependencies:
     - python=2
-    - pvl
-    - xml
     - lxml
-    - shutil
-    - pysis
     - pytz
-    - sqlalchemy
```
AustinSanders: We need sqlalchemy, not squalchemy (below)

eabovre: Oops! I'll fix that!

PR #66, opened 2018-06-04
Fixed duplicate primary key error, added --spread-job option to hpc jobs

AustinSanders:
+ small bug fixes


AustinSanders:
@USGSsakins The MetaBands table is [here](https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/dev/models/upc_models.py#L115).  It's also used in UPC_process and appears to be functioning properly.

PR #72, opened 2018-06-08
Initial rm_expired_pow

AustinSanders:


AustinSanders:
Everything is tested except the actual update / removal due to the fact that there's no clusterjobs dev database.

USGSsakins:
@AustinSanders, 

1.  we want the path to where the files are created/need to be removed from to be in a config file so that the cleanup script, POWprocess and MAPprocess all point at the same area.

2. for testing this, create a directory using a key from the dev database and populate it with files (even 0 length is okay) and set the timestamps in the dev database so that the directory needs to be cleaned up.

Issue #14, opened 2018-04-11
Run DIprocess.py inside of Airflow DAG


USGSsakins:
For the first pipeline to try out in Airflow use the DIprocess.py.  Break down this working solution as a Directed Acyclic Graph.  Figure out what makes sense for DAG steps (ie. calculate MD5 - then - update DB - ???)  Depends on #13

AustinSanders:
This is finished as of PR #75, but it would be cool (and probably pretty easy) to parallelize it by dynamically triggering subdags as shown [here](https://airflow.incubator.apache.org/faq.html#how-can-i-create-dags-dynamically).

AustinSanders:
It turns out it was pretty easy to get a parallel version up and running.  Since all tasks exist in the same subdag, the current implementation is somewhat of a synchronized-parallel implementation.  This means that n segments run in parallel, but iteration doesn't occur until the slowest segment is complete.  To get true parallelism, we'll have to spawn multiple subdags instead of relying on tasks within a single subdag.  

This is all due to the nature of DAGs and the inability to trigger tasks independently -- we have to trigger the entire dag or nothing at all.

Issue #13, opened 2018-04-11
Pipelines Docker Container


USGSsakins:
To allow for portability and easy upgrades we would like to have the pipeline framework in a Docker container.

Leverage this solution for airflow as an example https://github.com/puckel/docker-airflow?

USGSsakins:
Let's put all Docker files into a /services folder in the repo

AustinSanders:
@USGSsakins The docker file is actually giving me some trouble.  I'm having some trouble getting docker-compose to build an image that includes the local module.  As per [this](https://stackoverflow.com/questions/44708481/how-to-install-local-packages-using-pip-as-part-of-a-docker-build) post, it may be due to the directory structure (i.e. docker won't allow inclusion of local modules that are outside the directory containing the dockerfile).  It could also be due to the fact that I'm still fairly new to docker and may just be doing it wrong.

PR #75, opened 2018-06-29
Airflow integration

### LINKED ISSUE #15: Run DIqueue.py in AirFlow DAG
The second pipeline to migrate after getting DIprocess.py working is DIqueue.py. Create a DAG to figure out what files are due for processing and to schedule the DIprocess.py DAG for those files.  Depends on #14

AustinSanders:
*  Initial dockerfile + docker-compose.yaml pursuant to #13
*  Renamed 'bin' directory to 'pds_pipelines' to follow standard naming convention
*  Moved 'models' directory inside 'pds_pipelines' to allow for 'pds_pipelines.models.upc_models' style import.  This feels better (in my opinion) than a top level module named 'models'
*  Added initial di_process and di_queueing dags.
    *  di_queueing appears to be fully functional, needs error handling (closes #15)
    *  di_process appears to work up to and including file hashing (most of #14).  Needs to be run on astrovm4 for access to the products.


dags/tuto.py
```diff
@@ -0,0 +1,48 @@
+"""
```
USGSsakins: This looks like a canned AirFlow tutorial, do we need this in the PDS Pipelines project?

AustinSanders: Correct -- commit ea7eee1 involved renaming 'bin' directory to 'pds_pipelines.'  Otherwise, we'd have to import things as 'bin.models.upc_models' or similar.

AustinSanders: Commit 7db41bc removes tuto.py.  I think this was baked in with the airflow package that I downloaded and I didn't realize that it got pushed.

PR #102, opened 2018-08-10
Initial tool for generating json keyword skeleton from db

### LINKED ISSUE #100: Generate skeletons for keyword_desc file from keywords listed in database
Create a python program that will scrape the database for keywords associated with an archive and write a json skeleton similar to keyword_description.json.

Associated with #88

AustinSanders:
Closes #100


dpmayerUSGS:
@AustinSanders Please hang on to this code, but I don't think it's something we need to incorporate into the repo.

PR #124, opened 2018-08-29
Initial notification utility

AustinSanders:
Tested for successful + errant finishes.  Still need to add logic for upcoming purge notifications.  This should be pretty easy since ~90% of the code will be reused.  If we're running these on separate schedules, it may be easier to just create two separate scripts to handle "ready to download" and "ready to purge" notifications.

### DIFF
--- a/pds_pipelines/IngestProcess.py
+++ b/pds_pipelines/IngestProcess.py
@@ -5,6 +5,7 @@ import datetime
 import pytz
 import logging
 import json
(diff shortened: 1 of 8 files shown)


AustinSanders:
Refactored to avoid code duplication.  The logic for the notification processes is located in "notify.py" and the "notify_finished.py" and "notify_upcoming_purge.py" scripts serve as the executables.

PR #134, opened 2018-09-06
Factor out xml dependencies

### LINKED ISSUE #133: Eliminate XML dependencies
The LinkArtifacts script currently looks at the original XML recipes for the "file" and "datalink" tags.  In the spirit of consolidating recipes, we should move the relevant xml tags to JSON or to the config file.

AustinSanders:
Closes #133


USGSsakins:
Looks good

Issue #139, opened 2018-09-11
Uncaught Exception When Passing Bogus Archive Name to Ingestqueueing.py


dpmayerUSGS:
Normally, we would like `Ingestqueueing.py` to print a list of available archive names if the user passes an incorrect archive name. I just entered an incorrect archive name (`./Ingestqueueing -a jdhf`), and after the script printed the list of valid archives (as expected), it also printed this uncaught exception:

```
Traceback (most recent call last):
  File "./Ingestqueueing.py", line 75, in main
    archivepath = PDSinfoDICT[args.archive]['path'][:-1]
KeyError: 'jdhf'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "./Ingestqueueing.py", line 124, in <module>
    sys.exit(main())
  File "./Ingestqueueing.py", line 81, in main
    logging.error("Unable to locate {}".format(archivepath))
UnboundLocalError: local variable 'archivepath' referenced before assignment
```

I think https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/master/pds_pipelines/Ingestqueueing.py#L81 is supposed to reference `args.archive` rather than `archivepath`

PR #140, opened 2018-09-11
Fix Issue #139

eabovre:
In https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/18249d9fb80983559bbd730cbda6d006d4a119f1/pds_pipelines/Ingestqueueing.py#L81 changed "archivepath" to "args.archive"


AustinSanders:
Looks good - I'm merging this into dev.

Issue #148, opened 2018-09-14
Dynamically build list of process arguments for HCPjobber.py


dpmayerUSGS:
Currently, `HPCjobber.py` doesn't rely on argparse to enforce which arguments to the process option are valid and the current help strings contain an incomplete list of what the valid arguments are. 

A simple way to correct this would be to use an argparse choices container to store a list of valid arguments (https://docs.python.org/3.6/library/argparse.html#choices).

Rather than manually maintaining this list, it would make more sense to dynamically build a list from the keys in `jobconfig.py` (di, upc, ingest, etc.)

- [ ] Dynamically build list of available arguments to `--process` option in `HPCjobber.py` based on jobconfig.py
- [ ] Edit the argparse help string for `--process` option to reflect changes above

PR #150, opened 2018-09-14
Issue #148

eabovre:
Created 'choices' option in args based on key values in jobconfig.py, also edited 'help' option to match

### DIFF
--- a/pds_pipelines/HPCjobber.py
+++ b/pds_pipelines/HPCjobber.py
@@ -27,7 +27,7 @@ class Args:
 
     def parse_args(self):
         choices = jobconfig.keys()
-        print (choices)
+
         parser = argparse.ArgumentParser(description='PDS HPC Job Submission')
         parser.add_argument('--process', '-p', dest="process", required=True,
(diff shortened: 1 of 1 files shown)


pds_pipelines/HPCjobber.py
```diff
@@ -26,11 +26,11 @@ def __init__(self):
         pass
 
     def parse_args(self):
+        choices = jobconfig.keys()
```
dpmayerUSGS: Let's change this to `choices = list(jobconfig.keys())` so the help statement can just format the list to a string.

eabovre: Okay, I have updated it, should I close this PR and open a new one?

dpmayerUSGS: Just commit your edits and push them. The commit will be added to the PR.

eabovre: Done!

dpmayerUSGS: Remember to change the help string to just use `.format(choices)` in the help string

Issue #52, opened 2018-05-10
Shut down switch in Redis


USGSsakins:
Using scancel to kill SLURM jobs can hang the database connections causing all following jobs to fail since they cannot get a valid database connection.  To send a signal to processes, we need to set a key in Redis letting the different jobs know it is okay to continue running, or that they should commit their current changes, close the database connections and exit.

- [x] Add Redis key (true/false) for each process type (ie. DI_Run, UPC_Run)
- [x] Add logic to the processing scripts (ie. DIprocess.py) to check this run key inside the while loop before collecting the next file from the Redis queue.  If the key is set to false, exit the while loop, clean up connections/files and exit safely
- [x] Provide a way to turn on and off process types from the command line

AustinSanders:
Implemented in #154

PR #154, opened 2018-09-21
Initial RedisLock.py

AustinSanders:
RedisLock is a prototype object that could be used in #52.  Instead of having independent Redis Keys (one per queue), the RedisLock object will serve as a single point of access locking hub.  It uses an underlying hash table where keys are the names of redis queues that will be locked / unlocked.  Values are simply "0" or "1" -- 0 being locked, 1 being unlocked.  Currently, we can lock/unlock individual queues or lock/unlock all queues at once, but it may be worthwhile to be able to control queues by namespace as well, e.g. RL.lock_all(prefix='test_queue').

It should be made clear that locking a queue doesn't actually affect the associated queues.  If a script doesn't honor the locking mechanism, then the RedisLock makes no difference.  The idea here is that when a script connects to a RedisQueue, it will also register in the RedisLock and observe the values on its own.  The syntax would be as follows:

```
# Connect to the main queue
rq_main = RedisQueue('UPC_ReadyQueue')
# Connect to the RedisLock
rq_lock = RedisQueue('processes')
# Register process name (including namespace) in RedisLock
rq_lock.add({rq_main.name: '1'})

# Continue to do stuff until queue is empty or until the queue is locked
while rq_main.QueueSize() > 0 and rq_lock.available(rq_main.name):
    do_stuff()
```

Upcoming commits will include a script for interacting with the object / underlying hash table so that we can lock / unlock specific queues via command line.  Current options are to manipulate locks via redis-cli or via the python interpreter.

### DIFF
--- a/pds_pipelines/DIprocess.py
+++ b/pds_pipelines/DIprocess.py
@@ -10,6 +10,7 @@ import json
 from ast import literal_eval
 
 from pds_pipelines.RedisQueue import RedisQueue
+from pds_pipelines.RedisLock import RedisLock
 from pds_pipelines.PDS_DBquery import *
 
 from sqlalchemy import *
@@ -38,16 +39,17 @@ def main():
     logger.info('Starting DI Process')
 
     try:
-        # ignores engine information
-        session, _ = db_connect(pds_db)
+        session, engine = db_connect(pds_db)
         logger.info('DataBase Connecton: Success')
     except:
         logger.error('DataBase Connection: Error')
 
     RQ = RedisQueue('DI_ReadyQueue')
+    RQ_lock = RedisLock('processes')
+    RQ_lock.add({RQ.id_name: '1'})
     index = 0
 
-    while int(RQ.QueueSize()) > 0:
+    while int(RQ.QueueSize()) > 0 and RQ_lock.available(RQ.id_name):
         item = literal_eval(RQ.QueueGet().decode("utf-8"))
         inputfile = item[0]
         archive = item[1]
@@ -87,6 +89,10 @@ def main():
         index = 1
     except:
         session.rollback()
+
+    # Close connection to database
+    session.close()
+    engine.dispose()
 
 
 if __name__ == "__main__":
--- a/pds_pipelines/DIprocess.py
+++ b/pds_pipelines/DIprocess.py
@@ -15,7 +15,7 @@ from pds_pipelines.PDS_DBquery import *
 
 from sqlalchemy import *
 from sqlalchemy.orm.util import *
-from pds_pipelines.config import pds_db, pds_log, pds_info
+from pds_pipelines.config import pds_db, pds_log, pds_info, lock_obj
 from pds_pipelines.db import db_connect
 from pds_pipelines.models.pds_models import Files
 
@@ -45,7 +45,7 @@ def main():
         logger.error('DataBase Connection: Error')
 
     RQ = RedisQueue('DI_ReadyQueue')
-    RQ_lock = RedisLock('processes')
+    RQ_lock = RedisLock(lock_obj)
     RQ_lock.add({RQ.id_name: '1'})
     index = 0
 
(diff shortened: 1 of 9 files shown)


jlaura:
I want to understand a bit more the problem that this seeks to solve. You have n-jobs running on the cluster and one of these jobs exits abnormally (due to a hardware or network failure, an uncaught logic error, or an scancel call). When this happens the connection to the database is left open. As the timeout on the database connection is long, the database then exhausts all available connections.  Is that accurate?

Have you all ensure that:

(1) The DB has a maximum reasonable number of concurrent connections
(2) SQLAlchemy is using the NullPool so that a SQLAlchemy is not doing connection pooling for you

Have you thought about:
(1) Turning down the timeout
(2) Using something like pgbouncer as an external pooler that will recycle connections quite quickly. Note that this will force a refactor in instances where you: <create connection> + <run work longer than the bouncer timeout> + <try to access the session that is not closed>.

jlaura:
Looking at the PR code -> this should be rebased now that your other changes are in so that this PR focuses exclusively on the RedisLock.

USGSsakins:
We need a way to stop processing, but so that it stops gracefully by finishing processing any files it has already removed from the Redis queue.  This is not simply about database handles.

AustinSanders:
@jlaura As far as I understand, #52 isn't intended to address external failures or abnormal exit conditions.  As such, this isn't intended (at least at the moment) to provide graceful failure, but rather a graceful shutdown.  As the current system is implemented, there isn't really any way to issue a stop command.  We've already experienced scenarios in which there is scheduled down time, but our jobs are still running.  In those scenarios, we've been clearing the Redis Queue manually every time we want to stop something.

Instead of clearing + repopulating the queue every time we want to stop processing, we can just manipulate the RedisLock (which probably needs a new name...).  Any processing script can look at its  handle in the hash map to see if it is allowed to continue processing.

AustinSanders:
> Looking at the PR code -> this should be rebased now that your other changes are in so that this PR focuses exclusively on the RedisLock.

I've tried
```
git rebase upstream/dev
```
within my local dev branch, but it says I'm up to date.  What am I doing wrong?

jlaura:
Okay, this makes more sense now. Does it make sense that this is a class then? Or could a function lpush (I am assuming that all other commands are rpushing) and then clear the queue, etc. Basically, a function to inject a poison pill.

USGSsakins:
@jlaura We don't want to clear the queue.  We just want to stop running processes in case we need to let IT work on the NFS shares or something like that.  We want the queues to stay like they are so we can pick up where we left off.

jlaura:
so gracefully stop the currently running jobs, keep the redis queue populated and push the 'paused (stopped)' jobs back to the front of the queue?

AustinSanders:
@jlaura Right now, our RedisQueues are just queues of [(filename1, archiveid1), (filename2, archiveid2),....].  Our processing scripts just continually pop off a tuple of (filename, archiveid) and run di, upc, thumbnail generation, etc.  When we issue a stop command, we're still going to finish processing the current file (they only take a few minutes each), we just won't start the next iteration.  I don't think we'll need to requeue anything or push anything back onto the queue.

AustinSanders:
There are a few justifications for making this an object oriented locking mechanism rather than a one-off issued function. 

- With an object, we're able to make the lock stateful.  It's possible to not only stop a running process, but to disallow further processing until we unlock it.  This might be useful for jobs that are issued externally (users can request a POW or MAP2 job).  Instead of commenting out the POW processing objective from the cron job, we can just allow the queue to build up without anything being processed.

- It allows for a convenient means of issuing a "stop all" for all running processes.  When a process runs, it registers itself with the lock object, so we can just say lock_obj.stop_all() and it will gracefully shutdown all running processes.  I'm not sure how this would be implemented via a standalone function.

jlaura:
@AustinSanders That makes total sense.

AustinSanders:
There are two pieces missing before #52 can be marked complete.
- A command line utility for easy lock/unlock/stop functionality
- Integration with POW/MAP2 queues.

AustinSanders:
@jlaura @USGSsakins @dpmayerUSGS Everything needed to satisfy #52 is now implemented and ready for PR.  I've removed the "do not merge" status from this PR -- it should be ready for review + merge.

PR #155, opened 2018-09-21
Changed RedisQueue to FIFO

AustinSanders:
Tried to rebase + force push to squash local history, but I apparently messed up and ended up pushing things that aren't ready for push.  I reverted those changes, but I can't figure out how to PR without the 4328fb8 , 2754562, and 6504a8f, which should already be in master.

I don't want to mess with it any more out of fear of messing it up.


pds_pipelines/RedisQueue.py
```diff
@@ -106,5 +106,6 @@ def Qfile2Qwork(self, popQ, pushQ):
         str
             item
         """
-        item = self.__db.rpoplpush(popQ, pushQ)
+        item = self.__db.lpop(popQ)
```
jlaura: Why is this double underscore? This forces name mangling on the attribute to make it 'extra-not-private private'.

AustinSanders: Honestly, I'm not really sure.  This is leftover from Bob's code and I didn't know if it should be changed.

If this is something that should be changed, do you mind opening an issue?

Issue #79, opened 2018-07-25
Need Function to ID VOLDESC.CAT That Contain >1 value for DATA_SET_ID


dpmayerUSGS:
This is tangentially related to #77. 
The VOLDESC.CAT file for HiRISE (https://hirise-pds.lpl.arizona.edu/PDS/VOLDESC.CAT) contains a list of more than 1 item as the value of `DATA_SET_ID`. The old Perl softlinking code made an attempt to identify such cases by splitting on a comma, but based on the softlinks that currently exist in our file system, it does not appear that the old Perl code worked as intended.

The practice of creating a list of of values for `DATA_SET_ID` rather than a single value in a VOLDESC seems rare, but we don't know definitively which instruments follow this practice. I found the HiRISE example almost by accident.

It would be useful to have a little Python function that uses the PVL library to parse the value of `DATA_SET_ID` in all of our VOLDESC.CAT files and generate a simple report of that tells us which instruments in our archive contain more than 1 data set in a single VOLDESC.CAT.

eabovre:
https://github.com/eabovre/PDS-Pipelines/blob/Issue79-1/Issue%20%2379.ipynb

dpmayerUSGS:
@eabovre Scott and I have developed some specific requirements we'd like to see met so we can merge this notebook into dev:

- [ ] Input: The function should be able to read in a list of either URLs that point to VOLDESC.CAT files or fully-qualified paths to a VOLDESC.CAT file on a local file system

- [ ] Create a "tools" directory at the top of the repo and `git mv` the notebook there

- [ ] Add documentation to the Jupyter notebook that describes the circumstances under which the code should be run , the inputs, and the outputs

dpmayerUSGS:
I have also confirmed that it *is* valid for the `DATA_SET_ID` keyword to contain a list. Page 13 of Appendix A of the PDS standards reference provides an example `VOLDESC.CAT` that contains a long list of values for the keyword:
https://pds.nasa.gov/datastandards/pds3/standards/sr/AppendixA.pdf

eabovre:
Tests were successful!

eabovre:
Now I think I've actually fixed everything. I ended up doing a return from ds_count and making a dictionary called "a" in main

PR #157, opened 2018-09-25
Issue #79 Finished

eabovre:
Added all desired functionality to tools/Vold.py


tools/Vold.py
```diff
@@ -0,0 +1,89 @@
...
+        filePath = open(args.textfile,'r')
+        lines = filePath.readlines()
+        length = len(lines)
+        vol_val = {}
+        
+        for n in range(length):
+            if 'https' in lines[n] or 'http' in lines[n] or 'ftp' in lines[n]:
+                voldesc = urllib.request.urlopen(lines[n])
+                voldescPvl = pvl.load(voldesc)
+                dataset_id = voldescPvl['VOLUME']['DATA_SET_ID']
```
dpmayerUSGS: Rather than having to repeat the same 4 lines in 4 places, can the core logic (getting 2 of the keys from the PVL and returning information based on how many DATA_SET_ID values there are) be put into its own function?

Something like `ds_count(voldescPvl)` perhaps?
(I'm not beholden to that name for the function, it's just an example)

PR #163, opened 2018-10-04
Remove double underscores

eabovre:
Corrected unnecessary double underscores


pds_pipelines/models/upc_models.py
```diff
@@ -32,7 +32,7 @@ def create_table(keytype, *args, **kwargs):
 
 
 class DataFiles(Base):
-    _tablename_ = 'datafiles'
+    ___tablename___ = 'datafiles'
```
AustinSanders: Now these are triple underscored :(

eabovre: Oh my word, why did I make this so hard! I just fixed it, please let me know if there is anything else I breezed over!

pds_pipelines/models/upc_models.py
```diff
@@ -81,7 +81,7 @@ class Keywords(Base):
 
 class Meta(object):
     # Enforce compound primary key constraint
-    __table_args__ = (PrimaryKeyConstraint('upcid', 'typeid'),)
+    _table_args_ = (PrimaryKeyConstraint('upcid', 'typeid'),)
```
AustinSanders: One last change -- table args still requires double underscores.

eabovre: I am actually laughing out loud about how many times I've said "all right, this has to be the last one"

PR #173, opened 2018-11-01
Reduce DI SLURM memory allocation to 256 MB

dpmayerUSGS:


dpmayerUSGS:
I'm going to break with the usual protocol and merge my own PR because 

1. The change is trivial

2. The content of this PR is the same as #169, except #169 *should* have been requesting to merge to master

3. @USGSsakins approved #169

PR #175, opened 2018-11-06
Remove requirements.txt

eabovre:
used git rm to remove the requirements.txt file. I am not sure why the other commits are in this pull request...


dpmayerUSGS:
@eabovre Looks like you might need to rebase. If we were to merge this, I think it would revert some changes that we want to keep in dev. I'm going to close this PR just to reduce the chances that we break anything and we can work on making sure your fork is up to date next time you're in.

PR #180, opened 2018-11-16
Merge dev into master

AustinSanders:
Clearly, this is a large PR that consists of multiple parts.  In the future, we should try to sync dev and master more frequently.

Of the 1100+ line additions, almost 500 of the lines are contained within the newly created files db_sync.py and vold.py -- both of which have been tested and are ready for production.

Of the remaining ~600 added lines, ~50 are fixes to the RedisLock.  Again, these are tested and ready for production.

This leave over 500 additions that span the remaining ~20 files.  These changes address two main goals -- improving the logging capabilities of the processing + queueing programs, and more general "housekeeping" tasks like general readability, factoring out unused imports, and pep8 compliance.  While these changes are not explicitly tested, they carry no meaningful risks.  In the event that something breaks, it will simply create a less-than-perfect log item or, in the worst case, result in a syntax error.  I've run each item through a syntax checker, and it looks like everything is ok.  Isolating and testing each element of the pipeline is essentially a high-effort low-reward tradeoff, so I'm recommending that we merge the files and address the minor concerns as they appear.


dpmayerUSGS:
This PR is definitely overdue and would have ideally been split up into several PRs, but we (mostly me) got sloppy and didn't follow up PRs into dev with corresponding PRs into master.
I would be comfortable merging the new `db_sync.py` and `vold.py` if they appeared in their own PR. Ditto the updates to the RedisLock functionality.

The untested logging code looks fine to me, and I agree with the assessment that there is low risk associated with merging them into master at this time. It will be easier to squash minor bugs as they arise.

Squashing and merging.

PR #184, opened 2018-12-11
Dashboard feed

USGSsakins:
Fixed reading redis key for NFS load and output of JSON


jlaura:
@dpmayerUSGS @USGSsakins I played around more with Tipboard last night. I am happy to give you all the code, but I am not going to deploy it because it only interacts restfully, so I would need to sit yet another service in between anything that passively exposes a REST api and Tipboard. Easier to just use a react or angular dashboard that actively polls. Let me know if you all want to the code.

PR #187, opened 2018-12-19
Updated powprocess, mapprocess, and servicejobber

### LINKED ISSUE #185: Add key and namespace options to ServiceJobber
To help with testing and adding features in the future, it would be helpful to be able to pass a key (for queuing up previous POW/MAP2 jobs) and a namespace field (to override the default Redis namespace of "queue").

- [ ] Add command line argument to pass a POW/MAP2 job key
- [ ] Add command line argument to pass a Redis queue namespace

### LINKED ISSUE #186: Matching POW jobs to recipes
When processing POW jobs, there is no instrument name that matches the recipe JSON files.  To work around this, the URLs for the files being processed by POW can be used to lookup the recipes using the https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/master/pds_pipelines/PDSinfo.json.

1. The URL needs to have the protocol, hostname, and `mission` stripped from the front of the URL and the filename stripped from the end.
2. Next, prepend `/pds_san/PDS_Archives/` to the front of the URL
3. the PDSinfo.json includes a `path` variable in each block, use this to find the matching path to lookup the recipe name

AustinSanders:
- Updated POW/MAP2 to work with python 3
- Factored out hard-coded paths
- Eliminated (one) lookup table for MAP2/POW recipes (Closes #186) .
- Updated servicejobber to allow for namespace + key specification (Closes #185)
- Removed star imports
- General debugging + python formatting changes
- Added / improved error handling throughout

### DIFF
--- a/pds_pipelines/Servicejobber.py
+++ b/pds_pipelines/Servicejobber.py
@@ -46,15 +46,39 @@ class jobXML(object):
         Searches the PDSinfo dict for the 'clean name' that matches the recipes.  This function essentially
         maps URL->file path->internally consistent name.
         """
+        # @TODO Fix after refactor
+        # NOTE: I know this is really, really bad.  We're kind of backed into a corner here,
+        #  and partial string matching on a value-based lookup of a nested dict is the temporary solution.
+
         # Get any file listed.  Assumes that all files are from the same instrument
         file_name = self.getFileList()[0]
         file_name = file_name.replace('http://pdsimage.wr.usgs.gov/Missions/', archive_base)
-        # Note: I know this is really, really bad.  We're kind of backed into a corner here,
-        #  and partial string matching on a value-based lookup of a nested dict is the temporary solution.
+
+        candidates = []
+
         for key in self.pds_info:
             if file_name.startswith(self.pds_info[key]['path']):
-                return key
-        raise(KeyError('No key found in PDSInfo dict for path {}'.format(file_name)))
+                candidates.append(key)
+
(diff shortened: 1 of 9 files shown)


pds_pipelines/Servicejobber.py
```diff
@@ -832,16 +900,15 @@ def main():
...
+    # Test for GDAL and add to recipe
     Oformat = xmlOBJ.getOutFormat()
     if Oformat == 'GeoTiff-BigTiff' or Oformat == 'GeoJPEG-2000' or Oformat == 'JPEG' or Oformat == 'PNG':
         if Oformat == 'GeoJPEG-2000':
             Oformat = 'JP2KAK'
         if Oformat == 'GeoTiff-BigTiff':
             Oformat = 'GTiff'
         GDALprocessOBJ = Process()
-#        GDALprocessOBJ.newProcess('/usgs/dev/contrib/bin/FWTools-linux-x86_64-3.0.3/bin_safe/gdal_translate')
         GDALprocessOBJ.newProcess('/usgs/apps/anaconda/bin/gdal_translate')
```
dpmayerUSGS: @AustinSanders Ultimately, we should factor out the hard-coded path and use whichever version of gdal_translate is available in the environment, but I understand that we still have some research regarding GDAL GeoJPEG-2000 support. 
For now, let's insert a TODO comment identifying this line as something we want to factor out so that we don't forget, or at least to help us identify that it's fragile.

USGSsakins: We'll do an issue for the comment addition.

PR #209, opened 2019-04-01
Log hotfix

### LINKED ISSUE #206: Servicejobber.py: Fails when queue is empty
When trying to run Servicejobber.py when the queue is empty we get this failure:
```
  File "./Servicejobber.py", line 1019, in <module>
    sys.exit(main())
  File "./Servicejobber.py", line 603, in main
    key = DBQO.jobKey()
  File "/home/pds2010mgr/anaconda3/envs/PDS-Services/lib/python3.6/site-packages/pds_pipelines/PDS_DBquery.py", line 28, in jobKey
    for key in queryOBJ:
TypeError: 'NoneType' object is not iterable

```

We should handle an empty queue gracefully.

### LINKED ISSUE #207: Finaljobber.py: Incorrect syntax in a log statement
`logging.info('Nothing Found in Final Queue')` should be `logger.info('Nothing Found in Final Queue')`

Apparently the first is valid because it wasn't throwing an error, but it is incorrect.

AustinSanders:
Fixes #207 and Fixes #206 

Minor changes that address bugs in logging / error handling.

### DIFF
--- a/pds_pipelines/FinalJobber.py
+++ b/pds_pipelines/FinalJobber.py
@@ -40,10 +40,10 @@ def main():
 
 #***************Look at Final queue for work************
     RQ_final = RedisQueue('FinalQueue')
-    logger.info("Reddis Queue: %s", RQ_final.id_name)
+    logger.debug("Reddis Queue: %s", RQ_final.id_name)
 
     if int(RQ_final.QueueSize()) == 0:
-        logger.info('Nothing Found in Final Queue')
+        logger.debug('Nothing Found in Final Queue')
     else:
         FKey = RQ_final.QueueGet()
         logger.info('Found %s in Final Queue', str(FKey))
--- a/pds_pipelines/FinalJobber.py
+++ b/pds_pipelines/FinalJobber.py
@@ -43,7 +43,7 @@ def main():
     logger.info("Reddis Queue: %s", RQ_final.id_name)
 
     if int(RQ_final.QueueSize()) == 0:
-        logging.info('Nothing Found in Final Queue')
+        logger.info('Nothing Found in Final Queue')
     else:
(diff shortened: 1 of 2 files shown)


pds_pipelines/FinalJobber.py
```diff
@@ -43,7 +43,7 @@ def main():
     logger.info("Reddis Queue: %s", RQ_final.id_name)
 
     if int(RQ_final.QueueSize()) == 0:
-        logging.info('Nothing Found in Final Queue')
+        logger.info('Nothing Found in Final Queue')
```
dpmayerUSGS: @AustinSanders Let's change the logger level on this line to `debug` as well.
I don't think we need this message printed to the log every 2 minutes unless we're testing someting.

pds_pipelines/FinalJobber.py
```diff
@@ -43,7 +43,7 @@ def main():
     logger.info("Reddis Queue: %s", RQ_final.id_name)
```
dpmayerUSGS: @AustinSanders Let's change the logger level on this line to `debug`.
I don't think we need this message printed to the log every 2 minutes unless we're testing someting.

pds_pipelines/Servicejobber.py
```diff
@@ -600,7 +600,11 @@ def main():
     DBQO = PDS_DBquery('JOBS')
     if key is None:
         # If no key is specified, grab the first key
-        key = DBQO.jobKey()
+        try:
```
dpmayerUSGS: @AustinSanders Let's add a `logger.debug()` statement when we catch the exception and a `logger.info()` statement that prints the name of the first key. My comment is motivated by a desire to mirror what we do when looking for keys at the top of `FinalJobber.py`

PR #233, opened 2019-04-18
Added conditional decode to Qfile2Qwork function

### LINKED ISSUE #232: Need to decode job key for logging
It looks like we aren't decoding the jobkey everywhere in `POWprocess.py`. The script fails immediately with the following error output:
```
  File "/usgs/cdev/PDS-Pipelines/pds_pipelines/POWprocess.py", line 351, in <module>
    sys.exit(main())
  File "/usgs/cdev/PDS-Pipelines/pds_pipelines/POWprocess.py", line 72, in main
    logger = logging.getLogger(key + '.' + basename)
TypeError: must be str, not bytes
```

AustinSanders:
Fixes #232

### DIFF
--- a/pds_pipelines/RedisQueue.py
+++ b/pds_pipelines/RedisQueue.py
@@ -118,4 +118,4 @@ class RedisQueue(object):
             item
         """
         item = self._db.rpoplpush(popQ, pushQ)
-        return item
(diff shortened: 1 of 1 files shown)


AustinSanders:
PR is now stale.  These changes are included as part of #235.

PR #241, opened 2019-05-09
Adds NFS tool

AustinSanders:
History was ~70 commits out of sync between eabovre:dev and USGS-Astrogeology:master.  We weren't able to merge without significant manual conflict resolution, so I followed the following steps:

```
git checkout master
git pull upstream master
git checkout -b nfs_tool
git fetch https://github.com/eabovre/PDS-Pipelines.git dev
git cherry-pick 10218c2
git cherry-pick 309ce63
git cherry-pick d16b562
git cherry-pick 70a3e17
git push origin nfs_tool
```
I initially tried pulling with:
```
git pull -s recursive -X ours https://github.com/eabovre/PDS-Pipelines.git dev
```
followed by a rebase, but it got ugly and I didn't really feel like dealing with it.

I think this is the cleanest and easiest solution, but I'm not too famliar with cherry-picking commits -- I'm not sure if there are side effects.

### DIFF
--- a/tools/nfs_verify.py
+++ b/tools/nfs_verify.py
@@ -26,9 +26,8 @@ def main():
     args.parse_args()
 
     exists = os.path.isfile(args.path)
-    print(exists)
 
-    if exists is True:
+    if exists:
         print('NFS Mount was found')
     else:
         time = datetime.now().timestamp()
@@ -43,5 +42,7 @@ def main():
         text = msg.as_string()
         server.sendmail('astroweb@usgs.gov', args.email, text)
         server.quit
+
+
 if __name__ == '__main__':
     main()
--- a/tools/nfs_verify.py
+++ b/tools/nfs_verify.py
@@ -1,4 +1,4 @@
-import os, sys
+import os
 import argparse
 import smtplib
 from email.mime.text import MIMEText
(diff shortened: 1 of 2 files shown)


tools/nfs_verify.py
```diff
@@ -0,0 +1,47 @@
...
+
+def main():
+    server = smtplib.SMTP()
+    server.connect()
+    server.ehlo()
+    args = Args()
+    args.parse_args()
+
+    exists = os.path.isfile(args.path)
+    print(exists)
```
AustinSanders: Was this used for debugging, or do we want to print it here in production?

tools/nfs_verify.py
```diff
@@ -0,0 +1,47 @@
...
+
+    if exists is True:
```
AustinSanders: We can just use "if exists:" here.

dpmayerUSGS: @AustinSanders Feel free to add a commit to change that

tools/nfs_verify.py
```diff
@@ -0,0 +1,48 @@
...
+        pass
+
+    def parse_args(self):
+        parser = argparse.ArgumentParser()
+        parser.add_argument('--path', '-p', nargs = '?', help='file path')
+        parser.add_argument('--email','-e', nargs='+',help='list of emails')
+        args = parser.parse_args()
+        self.path = args.path
+        self.email = args.email
+
```
jlaura: This entire class is cargo cult style.

```python
def parse_args():
    parser = argparse.ArgumentParser()
    parser.add_argument('--path', '-p', nargs = '?', help='file path')
    parser.add_argument('--email','-e', nargs='+',help='list of emails')
    return parser.parse_args()

if __name__ == '__main__':
    args = parse_args()
    print(args.path)
    print(args.email)
    # profit 
```

dpmayerUSGS: @jlaura #JohnFrum

PR #251, opened 2019-06-12
Revert "Removed args classes in favor of more traditional argparse us…

dpmayerUSGS:
Reverts USGS-Astrogeology/PDS-Pipelines#247

The original PR should not have been merged. 

The original PR properly removed the Args classes, but the `parse_args()` function needs to be called from `main()` in order to actually work. This is usually accomplished with a line like `args = parse_args()`.
Then the variables corresponding to the arguments become defined and available as something like `args.foo`.

### DIFF
--- a/pds_pipelines/BatchDI.py
+++ b/pds_pipelines/BatchDI.py
@@ -11,21 +11,28 @@ from pds_pipelines.db import db_connect
 from pds_pipelines.config import pds_info, pds_db, pds_log
 
 
-def parse_args():
-    parser = argparse.ArgumentParser(description="DI Process")
+class Args(object):
+    def __init__(self):
+        pass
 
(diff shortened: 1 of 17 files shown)


AustinSanders:
I'm not sure that I understand the problem here.  parse_args does not need to be called within main because the arguments are passed to main as:
```
if __name__ == "__main__":
    args = parse_args()
    sys.exit(main(**vars(args)))
```

This will unpack the arguments in 'args' and pass them to main so that they are accessible.  This is equivalent to:
```
if __name__ == "__main__":
    args = parse_args()
    main(upcid=args.upcid, archive=args.archive)
```

dpmayerUSGS:
:-!
You're right. I'm too used to being a script kiddie that I forgot about that. Will close.

PR #268, opened 2019-08-06
Added error message for incorrect volume specification

### LINKED ISSUE #265: DIqueueing.py Not Warning of Invalid Volume Names
`DIqueueing.py` doesn't do anything to verify that a volume name passed on the commandline is valid or not, but it probably should. By contrast, if a user passes an invalid archive name, the script errors out and prints a list of valid archive names, but if the user passes an invalid *volume name*, the script simply accepts it. The script will query the DI database for entries matching the invalid volume name within the given archive, and always return exactly 0 results, because the volume doesn't exist.

There should be some sort of check that verifies whether there even exists an entry in the DI database with a path name that includes the given volume string. If the string doesn't match, then at a minimum, a warning should be printed to the DI log file. Ideally, the script would notify the user right away that the given volume is not found (similar to how it notifies the user that they passed an invalid archive name), but it probably wouldn't be a good idea to try and print a list of valid volume names because the list could be thousands of items long.

### LINKED ISSUE #266: FindDI_Ready.py Not Warning of Invalid Volume Names
This is essentially the same problem as #265 , but with the `FindDI_ready` script.

If a user passes an invalid volume name to `FindDI_Ready.py`, it will report that DI is current for that (non-existent) volume. The script should have some sort of check built in to verify that the volume is valid and notify the user if it is not.

Example of what currently happens when passing an invalid volume name:
```
FindDI_Ready.py -a lro_lroc_edr -v lsdhfshl
> lro_lroc_edr
> Database Connection Success
> Volume lsdhfshl DI Current
```

AustinSanders:
Fixes #265  and fixes #266 

Checks for the existence of any file matching the volume descriptor.  If no files were found, the scripts print an error message and exit.

At the moment, it is not feasible to provide any specific guidance regarding which volumes are available.  That is, we can tell users that their volume doesn't exist, but we cannot recommend volumes that do exist.

### DIFF
--- a/pds_pipelines/DIqueueing.py
+++ b/pds_pipelines/DIqueueing.py
@@ -102,7 +102,7 @@ def main():
     logger.addHandler(logFileHandle)
 
     logger.info("DI Queue: %s", RQ.id_name)
-    
+
     logger.info('Starting %s DI Queueing', args.archive)
     if args.volume:
         logger.info('Queueing %s Volume', args.volume)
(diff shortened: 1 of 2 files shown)


AustinSanders:
As a side note -- nothing is printed to the log.  I wasn't sure that it made sense to catalogue user error.  This can easily be added if anyone feels otherwise.

AustinSanders:
Closing this PR in favor of PR'ing to master

PR #310, opened 2019-09-20
Revised UPC

### LINKED ISSUE #287: UPC_process.py: Is find_keyword duplicated?
This block of code/function may be present in the same state from another file, verify this is needed or reduce to a single instance of this function

https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/d88df281472f9065f193d7fb0dc63646ab57971e/pds_pipelines/UPC_process.py#L38

### LINKED ISSUE #288: UPC_process.py: Add doc strings to functions
Add doc strings to functions for generating API docs for the UPC_process.py program.  Remove unnecessary block comments inline where not needed.

### LINKED ISSUE #289: UPC_process.py: Put error queue name in config
The name of the Redis error queue should be in the config.py, this would allow for all programs to write to the same queue with a simple change if we want that model in the future.

https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/d88df281472f9065f193d7fb0dc63646ab57971e/pds_pipelines/UPC_process.py#L121

### LINKED ISSUE #290: UPC_process.py: Move print statements to debug logging
Some output statements should write to the "debug" level logging:

https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/d88df281472f9065f193d7fb0dc63646ab57971e/pds_pipelines/UPC_process.py#L148

### LINKED ISSUE #291: UPC_process.py: Read paths and URL defaults from config.py
The SAN path and URLs need to be read from the config.py for ease of maintenance as things "move" across servers:

https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/d88df281472f9065f193d7fb0dc63646ab57971e/pds_pipelines/UPC_process.py#L151-L153

### LINKED ISSUE #292: UPC_process.py: Status string fixes
There are 2 issues with how the `status` variable is set and referenced

- [x] Remove this unnecessary "reset" https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/d88df281472f9065f193d7fb0dc63646ab57971e/pds_pipelines/UPC_process.py#L210
- [x] Since we set the string ourselves the same everytime, we don't need to call `.lower()`

### LINKED ISSUE #293: UPC_process.py: Add inline comment for "fix" on labels
This block of code needs a short comment explaining why we are doing regex fixes on the labels:

https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/d88df281472f9065f193d7fb0dc63646ab57971e/pds_pipelines/UPC_process.py#L233-L242

### LINKED ISSUE #295: UPC_process.py: Need to lookup `datafiles` entries by edrsource
Need to look for existing `datafiles` rows by the edrsource field

https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/d88df281472f9065f193d7fb0dc63646ab57971e/pds_pipelines/UPC_process.py#L270-L279

If the record is found, overwrite existing values in case anything has changed with new processing.

### LINKED ISSUE #296: UPC_process.py: Look up for instruments needs to include spacecraft
When looking up the DB entry for an instrument, we need to use both the instrument name (instrumentid) and the spacecraft since many spacecraft has instruments called "NAC" that were different instruments.

https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/d88df281472f9065f193d7fb0dc63646ab57971e/pds_pipelines/UPC_process.py#L258-L263

### LINKED ISSUE #297: UPC_process.py: Need to set detached label field (when needed)
When creating or updating `datafiles` entries, the detached label field should be populated with any detached label (ie. ending in .lbl).  The getkeywords to determine if there is a detached label will show the image file the label references (or it will be the same filename as the label).  Both of these fields should be URLs.

https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/d88df281472f9065f193d7fb0dc63646ab57971e/pds_pipelines/UPC_process.py#L277

### LINKED ISSUE #298: UPC_process.py: Remove call out for band list
With the new database schema, there is no longer a table specifically for the list of bands for an image.  This block of code is therefore no longer needed.

https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/d88df281472f9065f193d7fb0dc63646ab57971e/pds_pipelines/UPC_process.py#L289-L300

And these blocks setting up variables for the above redundant code:
https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/d88df281472f9065f193d7fb0dc63646ab57971e/pds_pipelines/UPC_process.py#L213-L214
https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/d88df281472f9065f193d7fb0dc63646ab57971e/pds_pipelines/UPC_process.py#L219-L220

### LINKED ISSUE #299: UPC_process.py: If a searchterm keyword is missing, set to null
When getting the keywords from the label for the searchterms table, if there is a missing keyword, set the searchterm column to `null`

https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/d88df281472f9065f193d7fb0dc63646ab57971e/pds_pipelines/UPC_process.py#L309

### LINKED ISSUE #300: UPC_process.py: Should we keep intermediate files?
Should some of the intermediate files from UPC_process.py be kept for generating browse, thumbnail, and map projected products?  Should this be done with a command line argument?

https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/d88df281472f9065f193d7fb0dc63646ab57971e/pds_pipelines/UPC_process.py#L344-L345

### LINKED ISSUE #301: UPC_process.py: Error block does not create instrument/target if new entry
The error processing tries to look up target and instrument info from the DB, but does not have the logic from above to create these DB entries if this is the first time we are dealing with an instrument or target.

Should the logic to do this be moved to a function so it is the same for the main branch and error case?

https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/d88df281472f9065f193d7fb0dc63646ab57971e/pds_pipelines/UPC_process.py#L364-L372

### LINKED ISSUE #302: UPC_process.py: Remove "fake" geometries for errors
The lines setting the center point and footprint for errors is no longer needed with the new database schema.  Remove these lines:

https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/d88df281472f9065f193d7fb0dc63646ab57971e/pds_pipelines/UPC_process.py#L390-L391

### LINKED ISSUE #304: UPC_process.py: Merge error1_input and error2_input
Change these to both be `error_input` so it is more clear in the code that they are doing the same thing for different exceptions.

### LINKED ISSUE #305: UPC_process.py: Log why the error block exception was called
If there was a problem writing/creating the error entry, the `catch` should write a log message to `warn`:

https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/d88df281472f9065f193d7fb0dc63646ab57971e/pds_pipelines/UPC_process.py#L438-L439

### LINKED ISSUE #306: UPC_process.py: Log to `error` when exiting for failure
This logger entry should write to `error` instead of `info`:

https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/d88df281472f9065f193d7fb0dc63646ab57971e/pds_pipelines/UPC_process.py#L351

### LINKED ISSUE #307: UPC_process.py: Final log message?
The final logging message of the UPC_process.py should be updated.  Take out the "success" part and if available include the `edrsource` variable:

https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/d88df281472f9065f193d7fb0dc63646ab57971e/pds_pipelines/UPC_process.py#L457

### LINKED ISSUE #308: HPC_jobber.py: Add --no-run Option
It would be nice to have a `--no-run` option in `HPCjobber.py` like the one in `ServiceJobber.py`:
https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/d88df281472f9065f193d7fb0dc63646ab57971e/pds_pipelines/Servicejobber.py#L576-L578

This option would have the effect of writing out the .sbatch file for the requested process (ingest, DI, UPC, etc.), but not actually submitting the job to SLURM.

This would make it easier to test/validate the .sbatch file without submitting it to the cluster.

### LINKED ISSUE #309: HPCjobber.py: Pass Commandline Options to Process Scripts
**Description**
`HPCjobber.py` is used to build and execute .sbatch scripts that wrap various PDS processes, such as Ingest, DI, and UPC. The python scripts that get wrapped may have their own commandline arguments, but there currently isn't a convenient way to pass those arguments to the process script from `HPCjobber.py`.

For example, `IngestProcess.py` has an optional argument, `--override` that allows it to re-ingest files that already exist in the DI database. In order to enable this option from `HPCjobber.py`, the user must enter `--process ingest-override` as opposed to `--process ingest`, which runs the ingest script without the override option. This requires maintaining separate entries on `jobconfig.py` for `ingest` and `ingest-override.` These entries are identical *except* that the latter includes the `--override` option.

It would be much simpler to be able to run `HPCjobber.py --process ingest --override` and have the override option appended to the line in `HPCjobber.py` that builds the actual command being wrapped by the .sbatch script:
https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/d88df281472f9065f193d7fb0dc63646ab57971e/pds_pipelines/HPCjobber.py#L77

**Proposed Solution**
Support for passing options on to the wrapped scripts can be accomplished by adding an optional positional parameter into `HPCjobber.py`'s argument parser with `nargs=argparse.REMAINDER`
Docs here:
https://docs.python.org/3.6/library/argparse.html?highlight=REMAINDER#nargs

**Definition of Done**
 - [x] add optional argument to `HPCjobber.py` (perhaps called `process-args`) with `nargs=argparse.REMAINDER`
 - [x] append list of any remainder argguments to the command being wrapped by `HPCjobber.py`
 - [x] remove `ingest-override` process from `jobconfig.py`

AustinSanders:
This PR closes a lot of issues.  normally, something like this would be split into multiple PRs, but given the nature of the issues (many issues were a single line fix / renaming variables), it was easier to issue a single PR.

This does include one large issue (parallelization), which was developed concurrently with the other bugfixes.  Since development was concurrent and there were logical dependencies, I built the smaller bugfixes in as I revised the workflow from sequential to parallel.

Closes #287 
Closes #288 
Closes #289 
Closes #290 
Closes #291 
Closes #292 
Closes #293 
Closes #295 
Closes #296 
Closes #297 
Closes #298 
Closes #299 
Closes #300 
Closes #301 
Closes #302 
Closes #304 
Closes #305 
Closes #306 
Closes #307 
Closes #308 
Closes #309


pds_pipelines/UPC_process.py
```diff
@@ -130,331 +180,298 @@ def main():
...
 
     # Disconnect from db sessions
     pds_session.close()
     session.close()
     # Disconnect from the engines
     pds_engine.dispose()
     upc_engine.dispose()
-    logger.info("UPC processing exited successfully")
+
+    # Log the EDRsource is available, otherwise default to inputfile
```
USGSsakins: Comment no longer needed

PR #321, opened 2019-11-07
UPCqueueing.py: Fix variable name mismatch, define undefined variable

### LINKED ISSUE #319: UPCqueuing.py: Variable name mismatch
**Description**
UPCqueing.py throws the following error when run in all cases except `UPCqueing.py -h`
```
  File "UPCqueueing.py", line 123, in <module>
    sys.exit(main(**vars(args)))
  File "UPCqueueing.py", line 81, in main
    if args.query_filter:
AttributeError: 'Namespace' object has no attribute 'query_filter'
```
**Additional Context**
Following block is supposed to apply string filtering is the user passed `--search` commandline option. The option used to be named `query_filter` but somehow the variable didn't get renamed in the execution logic (nor did the `args.` get stripped off the front after we changed how arguments are parsed).
https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/95daa00c1f384f86e45e245d178a34ce1a52a8cd/pds_pipelines/UPCqueueing.py#L81-L83

I'm not sure how this got missed because it prevents any realistic user test from passing. I suspect it slipped by during the Great db_refactor-->master Merge of FY19.

**Proposed Solution**
Change `args.query_filter` to `search` to match variable name in arguments list.

### LINKED ISSUE #320: UPCqueuing.py: `size_free` undefined
**Description**
The variable `size_free` is supposed to be a number representing the number of bytes of storage space free on the disk where UPC processing will take place, but the variable is not defined in the code, so UPCqueuing.py fails when it reaches the following block:
https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/95daa00c1f384f86e45e245d178a34ce1a52a8cd/pds_pipelines/UPCqueueing.py#L93-L95

**Proposed Solution**
Insert `size_free = disk_usage(workarea)[2]` immediately before the `if` block above

dpmayerUSGS:
Fixes #319 
Fixes #320

### DIFF
--- a/pds_pipelines/UPCqueueing.py
+++ b/pds_pipelines/UPCqueueing.py
@@ -78,8 +78,8 @@ def main(archive, volume, search, log_level):
     else:
         qOBJ = session.query(Files).filter(Files.archiveid == archiveID,
                                            Files.upc_required == 't')
-    if args.query_filter:
-        qf = '%' + args.query_filter + '%'
+    if search:
+        qf = '%' + search + '%'
         qOBJ = qOBJ.filter(Files.filename.like(qf))
 
     path = PDSinfoDICT[args.archive]['path']
@@ -90,6 +90,7 @@ def main(archive, volume, search, log_level):
             fname = path + element.filename
             size += getsize(fname)
 
+        size_free = disk_usage(workarea)[2]
         if size >= (disk_usage_ratio * size_free ):
             logger.error("Unable to process %s: size %d exceeds %d",
                          args.volume, size, (size_free * disk_usage_ratio))
--- a/pds_pipelines/UPCqueueing.py
+++ b/pds_pipelines/UPCqueueing.py
@@ -90,7 +90,7 @@ def main(archive, volume, search, log_level):
             fname = path + element.filename
             size += getsize(fname)
 
-        size_free = disk_usage(workarea)[2]
+        size_free = disk_usage(workarea).free
         if size >= (disk_usage_ratio * size_free ):
             logger.error("Unable to process %s: size %d exceeds %d",
                          args.volume, size, (size_free * disk_usage_ratio))


pds_pipelines/UPCqueueing.py
```diff
@@ -78,8 +78,8 @@ def main(archive, volume, search, log_level):
     else:
         qOBJ = session.query(Files).filter(Files.archiveid == archiveID,
                                            Files.upc_required == 't')
-    if args.query_filter:
-        qf = '%' + args.query_filter + '%'
+    if search:
```
jlaura: The way that the main is set up, every time that you make a change to the parser, you will have to add/subtract an arg from the main. This is a weird coupling. If you remove the `vars` from the call to main, you can simply pass in `args` (I might rename to avoid confusion with `*args, **kwargs` we commonly see. Then you can retain the `args.VARIABLE` syntax that is canonical. This way, you function signature is not tightly coupled to your argument parsing.

pds_pipelines/UPCqueueing.py
```diff
@@ -90,6 +90,7 @@ def main(archive, volume, search, log_level):
             fname = path + element.filename
             size += getsize(fname)
 
+        size_free = disk_usage(workarea)[2]
```
jlaura: This is a named tuple (https://docs.python.org/3/library/shutil.html#shutil.disk_usage). Why not just use `disk_usage(workarea).free` I think that that makes the intent of the line more understandable for someone new to the code base.

dpmayerUSGS: Ah, I hadn't thought of that! Will change.

pds_pipelines/UPCqueueing.py
dpmayerUSGS: Thanks, that's very helpful to know.  
We've made changes repo-wide more than a few times about how we parse arguments. It's worlds better than it was even several months ago, but I recognize there's still room for improvement. I have opened #322 so that we can address this in the future.

PR #341, opened 2019-12-12
Renamed upctime to processdate in the search_terms table

kaitlyndlee:


tests/database_test.py
```diff
@@ -31,9 +31,9 @@ def main():
 
             p = Polygon([ul, ll, lr, ur, ul])
 
-            upc_time = datetime.datetime.now(pytz.utc).strftime("%Y-%m-%d %H:%M:%S")
-            db_input.append(upc_models.SearchTerms(upctime=upc_time,
-                                            starttime=upc_time,
+            date_processed = datetime.datetime.now(pytz.utc).strftime("%Y-%m-%d %H:%M:%S")
+            db_input.append(upc_models.SearchTerms(processdate=date_processed,
+                                            starttime=date_processed,
```
USGSsakins: The process date and start time are different keywords.  Process date is when the file was processed into the UPC, the start time comes from the image label.

kaitlyndlee: I was copying what the original test was doing. It does not look like a real image is used for this test? Should I set the starttime to something else?

USGSsakins: @kaitlyndlee Then it was wrong in the previous version too.  The start time should come from the  keyword in the image label, @dpmayerUSGS what's the label group and keyword we should be using for the start time?

USGSsakins: @kaitlyndlee Sorry, I didn't realize this was a test.  For a test, the same date for both fields is fine.

PR #345, opened 2019-12-16
First jab at creating a script to create jsons for the views

kaitlyndlee:
This script queries the db for the summary and stores a json to disk. 

Considerations:
Better name for the script?
The top level for all of the jsons is 'json_agg' by default, should we change this?
Where should we put the jsons? Right now they are being written to pds_pipelines/json.


dpmayerUSGS:
Closing because @kaitlyndlee will work in a separate branch for setting up Jenkins. When that work is complete, a new PR can be opened to merge it into master.

Issue #313, opened 2019-09-25
UPC_process.py: Move recipeJSON inside of class


USGSsakins:
The recipeJSON is being passed around as an argument, this should be changed to be a member variable of the class (https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/master/pds_pipelines/Recipe.py).

kaitlyndlee:
@USGSsakins Are you referring to lines 235-240 in UPC_Process.py:
```
        # @TODO refactor this logic.  We're using an object to find a path, returning it,
        #  then passing it back to the object so that the object can use it.
        recipeOBJ = Recipe()
        recipe_json = recipeOBJ.getRecipeJSON(archive)
        #recipe_json = recipeOBJ.getRecipeJSON(getMission(str(inputfile)))
        recipeOBJ.AddJsonFile(recipe_json, 'upc')
```

USGSsakins:
@kaitlyndlee This issue was created based on the comment @AustinSanders entered (included in your comment).  Please follow up with Austin on how he suggests we resolve this.  I think the original intention was to make the recipe_json a member variable of the recipeOBJ class.

PR #346, opened 2019-12-19
Refactored recipe so that only one method needs be called to add a mission json

kaitlyndlee:
Added  addMissionJson that combines the two method calls that each script was doing. Removed getRecipeJSON since no files are using it now. Left in AddJsonFile since ServiceJobber is using it.

closes https://github.com/USGS-Astrogeology/PDS-Pipelines/issues/313

### DIFF
--- a/pds_pipelines/Recipe.py
+++ b/pds_pipelines/Recipe.py
@@ -33,6 +33,7 @@ class Recipe(Process):
         Parameters
         ----------
         file : str
+        proc : str
         """
 
         # @TODO use 'with' statement for file read
--- a/pds_pipelines/Recipe.py
+++ b/pds_pipelines/Recipe.py
@@ -15,6 +15,7 @@ from pds_pipelines.config import recipe_base
 
 class Recipe(Process):
     """
+    
     Parameters
     ----------
     Process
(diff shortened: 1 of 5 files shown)


pds_pipelines/Recipe.py
```diff
@@ -47,30 +47,25 @@ def AddJsonFile(self, file, proc):
...
-        """
-        return self.recipe
-
-    def getRecipeJSON(self, mission):
+    def addMissionJson(self, mission, proc):
         """
         Parameters
         ----------
         mission : str
+        proc : str
```
acpaquette: Could the doc string be updated for addMission and AddJsonFile?

pds_pipelines/Recipe.py
```diff
@@ -47,30 +47,25 @@ def AddJsonFile(self, file, proc):
...
-        """
-        Returns
-        -------
-        list
-            self.recipe
-        """
-        return self.recipe
-
-    def getRecipeJSON(self, mission):
+    def addMissionJson(self, mission, proc):
```
acpaquette: Not to be a nudge but it almost seems like this function doesn't need to exist if all it's doing is handling paths. Are we using the AddJsonFile function anywhere else?

kaitlyndlee: Yes, we are using it in ServiceJobber.py.

acpaquette:
Could we update the doc strings to be more like the doc string in getPDSid in UPCprocess? Short description of the function and a short description of each argument

PR #349, opened 2019-12-19
Testing Update

### LINKED ISSUE #347: UPC database ORM Changes
There are relationships in the ORM that need to be properly defined when instantiating a database using the ORM.

- [x] Add foreign key constraints to SearchTerms and JsonKeywords tables for upcid
- [x] Add relationship definitions for Datafiles for SearchTerms and JsonKeywords
- [x] Add relationship definitions for Targets and Instruments in SearchTerms and DataFiles
- [X] Added ForeignKey and relationship for upcproductid (Unable to do this as the productid is not guaranteed to be unique and foreign keys must be unique)
- [X] A fresh Database can be instantiated through the ORM

acpaquette:
Initial testing and moving processing logic into separate functions

Closes #347


pds_pipelines/models/upc_models.py
```diff
@@ -66,6 +78,8 @@ class Instruments(BaseMixin, Base):
     mission = Column(String(256))
     spacecraft = Column(String(256))
     description = Column(String(256))
+    search_terms = relationship('SearchTerms', backref="instruments", uselist=False)
+    datafiles = relationship('DataFiles', backref="instruments", uselist=False)
```
kaitlyndlee: Do we need a json_keyworkds relationship?

pds_pipelines/models/upc_models.py
```diff
@@ -80,6 +94,8 @@ class Targets(BaseMixin, Base):
     baxisradius = Column(Float)
     caxisradius = Column(Float)
     description = Column(String(1024))
+    search_term = relationship('SearchTerms', backref="targets", uselist=False)
+    datafiles = relationship('DataFiles', backref="targets", uselist=False)
```
kaitlyndlee: Same as above.

pds_pipelines/models/upc_models.py
acpaquette: No since the JsonKeywords table doesn't use ids or values from the Instruments or Targets table. It only has upcid and the json_keywords column. Hence the relationship definition on the DataFiles table

pds_pipelines/tests/test_upc.py
```diff
@@ -0,0 +1,177 @@
...
+    assert model.Targets.__tablename__ in tables
+
+def test_search_terms_exists(tables):
+    assert model.SearchTerms.__tablename__ in tables
+
+def test_json_keywords_exists(tables):
+    assert model.JsonKeywords.__tablename__ in tables
+
+def test_target_insert(session, session_maker, pds_label):
+    pds_label = PVLModule({'TARGET_NAME': 'TEST TARGET'})
```
kaitlyndlee: Is there a reason why we are defining the pds_label fixture above and then redefining it in each individual test?

acpaquette: Good catch

acpaquette:
@kaitlyndlee Do you want to take one last look over before merging?

PR #354, opened 2019-12-20
Removed unused files, functions, and merged PDS_DBsessions and PDS_DBquery

kaitlyndlee:


AustinSanders:
Can we also remove projectionbrowse_process.py?

PR #357, opened 2019-12-20
DI Database Testing

acpaquette:
A few changes to the DI database in the ORM along with basic testing for the DI database. This PR also updates all sessions used in any PDS script to create a session after a sessionmaker object is returned from the db_Connect function.

### DIFF
--- a/pds_pipelines/db_sync.py
+++ b/pds_pipelines/db_sync.py
@@ -62,7 +62,8 @@ def main(user_args):
     statistics = user_args.statistics
 
     logging.basicConfig(level=loglevel)
-    source_session, _ = db_connect(upc_db)
+    source_session_maker, _ = db_connect(upc_db)
+    source_session = source_session_maker()
 
     upc_ids = get_upc_ids(source_session, date, instrument,
                           spacecraft, target)
@@ -77,7 +78,8 @@ def main(user_args):
     target_sessions = {}
 
     for target in targets:
-        session, _ = db_connect(target)
+        Session, _ = db_connect(target)
+        session = Session()
         target_sessions[target] = session
 
     instrument = get_instrument(source_session, instrument,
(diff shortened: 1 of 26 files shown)


kaitlyndlee:
Can you merge with master? There are a few files that got deleted in a recently merged PR that are being re-added in this PR.

pds_pipelines/db_sync.py
```diff
@@ -62,7 +62,8 @@ def main(user_args):
     statistics = user_args.statistics
 
     logging.basicConfig(level=loglevel)
-    source_session, _ = db_connect(upc_db)
+    source_session_maker, _ = db_connect(upc_db)
+    source_session = source_session_maker()
```
AustinSanders: This seems inconsistent with other changes.  Doesn't db_connect now return a Session object instead of a sessionmaker?

pds_pipelines/models/upc_models.py
```diff
@@ -162,5 +163,7 @@ class JsonKeywords(BaseMixin, Base):
     # If the table does not exist, this will create it. This is used in case a
     # user has manually dropped a table so that the project is not wrecked.
     Base.metadata.create_all(tables=[DataFiles.__table__,
```
AustinSanders: Is there a reason to specify the tables as an argument here?

pds_pipelines/thumbnail_process.py
```diff
@@ -127,8 +127,10 @@ def main():
 
     PDSinfoDICT = json.load(open(pds_info, 'r'))
 
-    pds_session, pds_session = db_connect(pds_db)
-    upc_session, upc_session = db_connect(upc_db)
+    pds_session_maker, pds_session = db_connect(pds_db)
```
AustinSanders: See previous comment.  Are these Session objects or SessionMaker ?

acpaquette: It should probably be made more consistent but it does return a session maker object that when called generates connections to the database as a session object. I have been using `Session` for any instance that only has a single session maker object and then being more explicit when there are more than one session maker like this case.

pds_pipelines/models/upc_models.py
acpaquette: This is a copy paste from autocnet that replaces the autocnet table names with our tables. I'm not sure if the tables need to be in such a specified list like they are.

PR #362, opened 2019-12-23
Get Function Testing

### LINKED ISSUE #351: UPC Function Testing
The  following functions need to be tested in UPC:
- [x] getPDSid
- [x] getISISid

acpaquette:
Testing for the two get functions (getISISid and getPDSid) inside of the UPCprocessing script.

Closes #351

### DIFF
--- a/pds_pipelines/tests/test_upc.py
+++ b/pds_pipelines/tests/test_upc.py
@@ -97,16 +97,16 @@ cam_info_dict = {'upcid': 1,
 
 @pytest.mark.parametrize('get_key_return', [('PRODUCTID'), (b'PRODUCTID')])
 @patch('pds_pipelines.UPCkeywords.UPCkeywords.__init__', return_value = None)
-def test_get_isis_id(mocked_init, get_key_return, pds_label):
+def test_get_pds_id(mocked_init, get_key_return, pds_label):
     with patch('pds_pipelines.UPCkeywords.UPCkeywords.getKeyword', return_value=get_key_return) as mocked_getkey:
         prod_id = getPDSid(pds_label)
     assert isinstance(prod_id, str)
     assert prod_id == 'PRODUCTID'
 
-def test_get_pds_id():
+def test_get_isis_id():
     cube_path = '/Path/to/my/cube.cub'
     serial = getISISid(cube_path)
-    serial == 'ISISSERIAL'
+    assert serial == 'ISISSERIAL'
     assert isinstance(serial, str)
 
 def test_datafiles_exists(tables):
--- a/pds_pipelines/tests/test_upc.py
+++ b/pds_pipelines/tests/test_upc.py
@@ -11,6 +11,14 @@ import numpy as np
 from unittest.mock import patch, PropertyMock
 from unittest import mock
 
+import pysis
+from pysis import isis
+
+# Stub in the getsn function for testing
+def getsn(from_):
+    return b'ISISSERIAL'
+pysis.isis.getsn = getsn
+
 import pds_pipelines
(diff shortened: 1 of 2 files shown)


pds_pipelines/tests/test_upc.py
```diff
@@ -87,6 +95,20 @@ def pds_label():
                                            (153.30256122853893, -32.68515128444211),
                                            (153.80256122853893, -32.68515128444211)]).wkt}
 
+@pytest.mark.parametrize('get_key_return', [('PRODUCTID'), (b'PRODUCTID')])
+@patch('pds_pipelines.UPCkeywords.UPCkeywords.__init__', return_value = None)
+def test_get_isis_id(mocked_init, get_key_return, pds_label):
```
kaitlyndlee: Swap the test names. getISISid is in pds test.

kaitlyndlee: Swap the test names. getISISid is in pds test.

pds_pipelines/tests/test_upc.py
```diff
@@ -87,6 +95,20 @@ def pds_label():
...
+    with patch('pds_pipelines.UPCkeywords.UPCkeywords.getKeyword', return_value=get_key_return) as mocked_getkey:
+        prod_id = getPDSid(pds_label)
+    assert isinstance(prod_id, str)
+    assert prod_id == 'PRODUCTID'
+
+def test_get_pds_id():
+    cube_path = '/Path/to/my/cube.cub'
+    serial = getISISid(cube_path)
+    serial == 'ISISSERIAL'
```
kaitlyndlee: missing assert?

pds_pipelines/tests/test_upc.py
```diff
@@ -11,6 +11,14 @@
 from unittest.mock import patch, PropertyMock
 from unittest import mock
 
+import pysis
+from pysis import isis
+
+# Stub in the getsn function for testing
```
kaitlyndlee: Are you not able to add this as a patch instead of redefining the function this way?

pds_pipelines/tests/test_upc.py
```diff
@@ -87,6 +95,20 @@ def pds_label():
                                            (153.30256122853893, -32.68515128444211),
                                            (153.80256122853893, -32.68515128444211)]).wkt}
 
+@pytest.mark.parametrize('get_key_return', [('PRODUCTID'), (b'PRODUCTID')])
+@patch('pds_pipelines.UPCkeywords.UPCkeywords.__init__', return_value = None)
+def test_get_isis_id(mocked_init, get_key_return, pds_label):
+    with patch('pds_pipelines.UPCkeywords.UPCkeywords.getKeyword', return_value=get_key_return) as mocked_getkey:
+        prod_id = getPDSid(pds_label)
+    assert isinstance(prod_id, str)
```
kaitlyndlee: Asserts need to be in with block?

pds_pipelines/tests/test_upc.py
acpaquette: This has to be like this due to how pysis import uses the ISIS binarys to generate application calls on the module. It collects the binaries then exposes them on the isis module if it finds them in the bin. Since there is not bin, the isis module has no isis functions attached to it so it can't be mocked the same way we have been mocking other functions/attributes

pds_pipelines/tests/test_upc.py
acpaquette: The return from the function should be able to exit the with context without any problems. Checking if the function was called with the appropriate stuff would need to be in the with block.

PR #364, opened 2019-12-31
Queue Testing and Function Splits

acpaquette:
There's a lot going on here:
- Small fix for Ingestprocess.py
- Splits the for loops up in the main UPCProcess.py script
    - Converts them into two seperate functions. One that handles reading and generating an isis process dictionary and another that executes said process dictionary.
- Adds test for the basic redis queue functionality of pushing and popping an item from the queue
- Tests the one of the newly split functions (Still need to add the second)


acpaquette:
Currently failing due to the config. Needs to be changed to a local path to get whatever json file is needed.

pds_pipelines/PDSinfo.json
```diff
@@ -44,7 +44,8 @@
     "galileo_ssi_edr": {
         "archiveid": "25",
         "path": "/pds_san/PDS_Archive/Galileo/SSI/",
-        "bandbinQuery": "FilterName"
+        "bandbinQuery": "FilterName",
+        "upc_reqs": ["/SSI/", ".lbl", ["c3", "g1", "g2", "j0"]]
```
dpmayerUSGS: `"/SSI"` should be removed.
`["c3", "g1", "g2", "j0"]` must be removed or else most SSI data won't process

PR #368, opened 2020-01-03
DO NOT MERGE: Removed thmproc-specific logic. Moved even/odd to handmos branch.

### LINKED ISSUE #325: UPC_process.py: Deal Generically with Odd/Even Detectors
**Description**
`UPC_process.py` contains messy logic for dealing specifically with THEMIS processing. For example:
https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/2cf655230a20899483b3885cb8a3bfd5cfc171ba/pds_pipelines/UPC_process.py#L241-L251

Much of this is motivated by the need to deal with separate files for the even and odd-numbered detectors from the THEMIS VIS instrument. Currently, THEMIS VIS is the only UPC-eligible instrument that gets split into odd/even detectors during processing, but we expect to be (re-)adding UPC support for MARCI in the near future, which also splits output between odd/even detectors after `marci2isis`. As such, it would be nice if there was a more generic way to handle odd/even detector cubes during UPC processing without having to resort to instrument-specific logic in `UPC_process.py`.

I'm not sure if this would work, but perhaps logic for dealing with odd/even detectors could be shifted into the recipe JSON for affected instruments?

AustinSanders:
Removed the thmproc branch from the generate_isis_processes so that it defaults to
general process logic.  Moved even/odd path calculation to handmos branch.

These changes are necessary in order to allow for a more generalized workflow that
supports additional instruments that use even/odd detectors (e.g. MARCI).

Closes #325

### DIFF
--- a/pds_pipelines/upc_process.py
+++ b/pds_pipelines/upc_process.py
@@ -435,17 +435,13 @@ def generate_isis_processes(inputfile, archive, logger):
         if '2isis' in item:
             processOBJ.updateParameter('from_', inputfile)
             processOBJ.updateParameter('to', outfile)
-        elif item == 'thmproc':
-            processOBJ.updateParameter('from_', inputfile)
-            processOBJ.updateParameter('to', outfile)
-            thmproc_odd = str(workarea) + str(os.path.splitext(
-                os.path.basename(inputfile))[0]) + '.UPCoutput.raw.odd.cub'
-            thmproc_even = str(workarea) + str(
-                os.path.splitext(os.path.basename(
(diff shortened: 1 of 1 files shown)


acpaquette:
Do we want to try running a marci or themis image through the pipeline? I don't know if we have the time but might be worth something to see where things break

AustinSanders:
As of 68a05e0 themis vis works with UPC.

AustinSanders:
This PR also adds back in some of the weird input / output renaming that has always been part of the pipeline.  We should consider finding a new / better solution.

AustinSanders:
This PR and its related work / tickets are obsolete.

PR #373, opened 2020-01-16
Added gdal to environment. Increase DI walltime to 15 days. Fix permi…

dpmayerUSGS:
…ssions on ingest_queueing.py


thareUSGS:
I'm confused, I thought GDAL and ISIS3 can't live in the same env.? I assume ISIS is there since pysis is in the env. (or is there an older ISIS being used)?

dpmayerUSGS:
@thareUSGS Good question.

We deploy and use PDS-Pipelines in such a way that gdal and ISIS *don't* live in the same environment. We have a conda environment that contains gdal (among other things) but not ISIS. Instead, we just make sure that `$ISISROOT` is defined in our `.bashrc` file, which allows `pysis` to find and use ISIS programs as needed.  We've been doing this successfully for over a year. 

Historically, we had built and managed the conda environment manually and weren't actually using the `environment.yml` file in this repo. During the last set of sprints, the devs cleaned up the `environment.yml` file to reflect what we actually use in the production environment, but accidentally left out gdal, so this PR just adds it in. This means that someone starting from scratch can create a conda environment suitable for running PDS-Pipelines by using the `environment.yml` file, minus ISIS. ISIS needs to be installed elsewhere (in another conda environment or otherwise), and then `$ISISROOT` defined in the user's `.bashrc`.

Issue #374, opened 2020-02-11
upc_process.py: File naming/renaming logic is broken


dpmayerUSGS:
### Description ###
The current logic for naming input/output files and passing them to particular conditions inside the big if/elif/else block in `generate_isis_processes()` is broken. The function names the output of `*2isis` or `thmproc` as `outfile`, but then expects the input to functions that are run subsequently in the recipe (such as `spiceinit` or `footprintinit`) as `infile`. This prevents the functions from building a correctly-parameterized set of processes from a given recipe because the variables `infile` and `outfile` have different values. 

https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/93d268bd0aec55057798d78c37068d958232a981/pds_pipelines/upc_process.py#L416-L467

#### Quick Fix ####
A short-term fix would be to just change the value of the "to" parameter in `*2isis` and `thmproc` to `infile`, redefine `infile` in the `handmos` case to have the same value as `thmproc_odd`, change the value of the "from" parameter in the `spiceinit`, `footprintinit`, and `caminfo` cases to `infile`, and remove the `cubeatt` case altogether. None of the recipes for high-priority datatsets (list in #281) actually *need* `cubeatt`, and it just complicates the logic in this block.

#### Long-term Solution ####
One way to make the function more general and robust to arbitrary recipes would be to implement a system of find/replace tokens in the recipe files (kind of like Jinja templates). This could eliminate the need for the if/elif/else block altogether, by allowing the individual recipes to guide the expected names of input/output files to each process. In practice, the only element of an input/output filename that can't be predicted in advance is the basename of the files. It should be possible for `upc_process.py` to insert the basename of a given product into the "to" or "from" value in a particular ISIS command, with the suffix defined within the recipe itself. This would render #325 moot.

Implementing this solution will require existing recipes to be edited/reformatted to be compatible with the templating system.

jlaura:
![](https://github.trello.services/images/mini-trello-icon.png) [#374](https://trello.com/c/JnBKjs35/266-374)

Issue #281, opened 2019-08-12
Test UPC-Eligible Datasets in Refactored UPC


dpmayerUSGS:
**Is your feature request related to a problem? Please describe.**
This issue is just a place to keep track of the status of which UPC-eligible datasets have been tested in the refactored UPC.

**Describe the solution you'd like**
All datasets that are currently eligible for UPC should run correctly under the refactored UPC system.

**Definition of done**
The datasets in the checklist below should all work correctly in the refactored UPC for this issue to be considered done. The names in the list are taken from `PDSinfo.json` and may change in the future. 
 - [ ] mro_ctx
 - [ ] lro_lroc_edr
 - [ ] galileo_ssi_edr
 - [ ] cassini_iss
 - [ ] mgs_moc
 - [ ] mo_themis_vis_edr
 - [ ] mo_themis_ir_edr
 - [ ] messenger
 - [ ] mariner_10
 - [ ] viking_orbiter
 - [ ] voyager

**Additional notes**
Separate issues should be opened to address datasets that fail testing in the refactored UPC. Those issues should reference this one.

dpmayerUSGS:
I removed Mariner 10 from the list because this isn't a dataset that was traditionally supported in UPC. I'm not sure why I put that on there originally.

PR #378, opened 2020-02-12
UPC Recipe Validation and upc_process bugfixes

### LINKED ISSUE #375: upc_keywords.py: Missing import statement
The UPCKeywords class relies on python's built in `re` library when reading poorly-formatted PDS labels, but the corresponding import statement is missing from `upc_keywords.py`, thereby causing `UPCkeywords()` to break when given poorly-formatted PDS labels.

## Solution ##
Insert `import re`

### LINKED ISSUE #376: upc_process.py: Handle .QUB and .imq Files When Checking for Detached Labels
### Description ###
`create_datafiles_record()` attempts to determine whether a given PDS file has an attached or detached label by parsing the value of the PDS3 pointer `^IMAGE`. If the value of the pointer is not an array, then the script infers that the product does not have a detached label. 
https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/93d268bd0aec55057798d78c37068d958232a981/pds_pipelines/upc_process.py#L236-L238

However, certain PDS3 products, such as THEMIS spectral cubes and some legacy products that are stored in a compressed format (such as MOC EDRs) do not contain this pointer, which causes this function to raise a KeyError and the script errors out. 

### Proposed Solution ###
In practice, the types of PDS3 files that do not contain this pointer have attached labels. Thus, changing the exception condition to include both `TypeError` and `KeyError` should correctly identify files that do not contain `^IMAGE` has having attached labels.

Fixing this issue is a prerequisite for closing #281

### LINKED ISSUE #377: upc_process.py: Handle products missing INSTRUMENT_NAME
Some datasets such as MOC EDRs are missing `INSTRUMENT_NAME` from their labels, even though it is a [required keyword](https://pds.nasa.gov/datastandards/pds3/standards/sr/Chapter05.pdf). For these cases, `upc_process.py` will fail to extract a sensible instrument name.

https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/93d268bd0aec55057798d78c37068d958232a981/pds_pipelines/upc_process.py#L172-L175

Even though a label may be missing the required `INSTRUMENT_NAME` keyword, it may alternatively contain the `INSTRUMENT_ID` keyword (as is the case with MOC EDRs).

We should implement a for loop with a break condition to look for keywords that might reasonably contain information about an instrument's name, much as we do for the name of the spacecraft:

https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/93d268bd0aec55057798d78c37068d958232a981/pds_pipelines/upc_process.py#L176-L189

dpmayerUSGS:
This PR includes small tweaks to UPC recipes for several datasets to make the recipes valid, and larger changes to `upc_process.py` to enable it to support recipes from a wider variety of datasets. The changes in this PR have been tested on products from all of the datasets listed in #281 except MESSENGER, Viking Orbiter, and Voyager. 

Fixes #376 
Fixes #375 
Implements suggested quick fix for #374 
Fixes #377


pds_pipelines/upc_process.py
```diff
@@ -462,7 +476,6 @@ def generate_isis_processes(inputfile, archive, logger):
             processOBJ.updateParameter('to', outfile)
```
AustinSanders: This is now the only reference to outfile.   I'm not sure that I fully understand how we're using infile / inputfile / outfile at this point.  It seems like inputfile is the original file and infile is both the output of a process _and_ the input to the next process, but what is outfile?  How does its usage differ from infile?

dpmayerUSGS: Yes, I think `outfile` will go away eventually. `outfile` was never really being used for anything because for *most* recipes, other than calling the initial `*2isis` program, the other ISIS programs modify a file in place (i.e., `spiceinit`, `footprintinit`).

My expectation is that this entire if/elif block can be factored out by moving some find/replace tokens for the input/output filenames into the recipes themselves. See #374

PR #388, opened 2020-04-15
Recipe Rendering with Jinja Templates

### LINKED ISSUE #374: upc_process.py: File naming/renaming logic is broken

acpaquette:
Changes recipe construction/processing to use jinja templates. 

Any recipe template is filled in with given values from the script, currently using the following:
- inputfile (Original source image ie. QUB, img, IMG, etc.)
- no_extension_inputfile (Original source image without the extension)
- cam_info_file (Output file to be used for cam info)

This set of parameters puts more onus on recipe creators to handle proper inputs and outputs and to know exactly how their recipe/pipeline works. In this PR are two updated pipelines for themis_vis and messenger. The themis_vis pipeline is a tested and true pipeline but exemplifies why someone needs to understand the pipeline as it uses an intermediate output from thmproc to run in the rest of the pipeline.

Closes #374

### DIFF
--- a/pds_pipelines/tests/test_process.py
+++ b/pds_pipelines/tests/test_process.py
@@ -1,0 +1,103 @@
+from pds_pipelines.process import Process
+
+import json
+
+import pytest
+from unittest.mock import patch, PropertyMock, Mock
+from unittest import mock
+
+@pytest.fixture
+def empty_process_obj():
+    process = Process()
+    return process
+
+@pytest.fixture
+def process_obj():
+    process = Process()
+    json_dict = {'data2isis': {'from_':'some_image.img', 'to': 'some_image.cub'},
+                 'spiceinit': {'from_':'some_image.cub'}}
+    json_str = json.dumps(json_dict)
+    process.JSON2Process(json_str)
+    return process
+
+@pytest.fixture
+def json_dict():
+    json_dict = {'data2isis': {'from_':'some_image.img', 'to': 'some_image.cub'},
+                 'spiceinit': {'from_':'some_image.cub'}}
+    return json_dict
+
+def test_default_process(empty_process_obj):
+    assert len(empty_process_obj.getProcess().keys()) == 1
+    assert empty_process_obj.getProcess()[''] == None
+    assert empty_process_obj.getProcessName() == ''
+
+def test_process_2_json(process_obj, json_dict):
+    process_str = process_obj.Process2JSON()
+    assert process_str == json.dumps(json_dict)
+
+# This seems "wrong"
+def test_json_2_process(empty_process_obj, json_dict):
+    json_str = json.dumps(json_dict)
+    empty_process_obj.JSON2Process(json_str)
+    assert empty_process_obj.getProcess().keys() == json_dict.keys()
+    assert empty_process_obj.getProcessName() == 'spiceinit'
+
+def test_set_process(process_obj):
+    assert process_obj.getProcessName() == 'spiceinit'
+    process_obj.setProcess('Banana')
+    assert process_obj.getProcessName() == 'Banana'
+    assert len(process_obj.getProcess().keys()) == 2
+
+# Reduces the number of processes in the process object from 2 to 1
+# Also seems "wrong"
+def test_change_process(process_obj):
+    assert process_obj.getProcessName() == 'spiceinit'
+    assert len(process_obj.getProcess()) == 2
+    process_obj.ChangeProcess('Banana')
+    assert len(process_obj.getProcess()) == 1
+    assert process_obj.getProcessName() == 'Banana'
+    assert list(process_obj.getProcess().keys()) == ['Banana']
+    assert list(process_obj.getProcess()['Banana'].keys()) == ['from_']
+
+# Reduces the number of processes from the json dict from 2 to 1
+# Also seems "wrong"
+def test_process_from_recipe(empty_process_obj, json_dict):
+    recipe = [{i: json_dict[i]} for i in json_dict]
+    empty_process_obj.ProcessFromRecipe('spiceinit', recipe)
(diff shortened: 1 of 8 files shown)


pds_pipelines/upc_process.py
```diff
@@ -590,9 +560,9 @@ def main(user_args):
         AddProcessDB(pds_session, fid, True)
         pds_session.close()
 
-        if not persist:
-            os.remove(infile)
-            os.remove(caminfoOUT)
+        # if not persist:
```
AustinSanders, line 563: Should these stay commented out?

pds_pipelines/tests/test_process.py
```diff
@@ -0,0 +1,103 @@
...
+def test_default_process(empty_process_obj):
+    assert len(empty_process_obj.getProcess().keys()) == 1
+    assert empty_process_obj.getProcess()[''] == None
+    assert empty_process_obj.getProcessName() == ''
+
+def test_process_2_json(process_obj, json_dict):
+    process_str = process_obj.Process2JSON()
+    assert process_str == json.dumps(json_dict)
+
+# This seems "wrong"
```
kaitlyndlee, line 38: For the things that you say seem "wrong", is this something that you think the code shouldn't do?

pds_pipelines/upc_process.py
```diff
@@ -567,19 +537,19 @@ def main(user_args):
         # Update the logger context to include inputfile
         context['inputfile'] = inputfile
 
-        processes, infile, caminfoOUT, pwd = generate_isis_processes(inputfile, archive, logger)
-        failing_command = process_isis(processes, workarea, pwd, logger)
+        processes, infile, caminfoOUT, pwd = generate_processes(inputfile, archive, logger)
+        failing_command = process(processes, workarea, pwd, logger)
```
kaitlyndlee, line 541: Not your change, just curious why we call it failing_command?

pds_pipelines/upc_process.py
kaitlyndlee, line 563: Can we remove the commented lines?

pds_pipelines/upc_process.py
```diff
@@ -567,19 +537,19 @@ def main(user_args):
...
         ######## Generate DataFiles Record ########
         upc_id = create_datafiles_record(pds_label, edr_source, infile, upc_session_maker)
- 
+
         ######## Generate SearchTerms Record ########
         create_search_terms_record(pds_label, caminfoOUT, upc_id, infile, upc_session_maker)
 
         ######## Generate JsonKeywords Record ########
-        create_json_keywords_record(caminfoOUT, upc_id, inputfile, failing_command, upc_session_maker)
+        create_json_keywords_record(caminfoOUT, upc_id, inputfile, failing_command, upc_session_maker, logger)
```
kaitlyndlee, line 552: Should we add the logger to the other functions? Can make a new issue if it's something we should add.

pds_pipelines/upc_process.py
acpaquette: No but the logic for only removing the files generated from the process and not the original file (QUB, IMG, img etc.) is a little more complicated. It's more all or nothing now but I will work on a solution that maintains the current functionality.

pds_pipelines/tests/test_process.py
acpaquette: Yes, from the original design of these classes it seems like they should only be able to contain one process but the JSON2Process allows for multiple "processes" to exist in one Process object. Further more the Process object seems somewhat counter intuitive when the desired behavior of the Process object can be accomplished by a dictionary. At least this is how I see it with the current state of the recipes since they aren't too complex and only deal with isis applications.

pds_pipelines/upc_process.py
acpaquette: I replied to Austin with an explanation of why this was commented out. I will work on a solution that will accomplish the same thing.

pds_pipelines/upc_process.py
acpaquette: I would like to make the logger more global so that it doesn't need to be passed around everywhere through function signatures. Not exactly sure how to do that but something to think about.

PR #389, opened 2020-04-15
Fixes session maker vs session problem

### LINKED ISSUE #372: service_job_manager.py: Fails when called with no arguments
**Description**
`service_job_manager.py` is typically called without any arguments, in which case it simply grabs the first key it sees in the job queue. However, it is now failing with the following error:
```
  File "pds_pipelines/service_job_manager.py", line 1036, in <module>
    sys.exit(main(parse_args()))
  File "pds_pipelines/service_job_manager.py", line 608, in main
    key = DBQO.jobKey()
  File "/home/pds2010mgr/anaconda3/envs/PDS-Services/lib/python3.7/site-packages/pds_pipelines/pds_db_query.py", line 67, in jobKey
    queryOBJ = self.session.query(self.processingTAB.key).filter(
AttributeError: 'sessionmaker' object has no attribute 'query'
```

**Steps to Reproduce**
`service_job_manager.py`

acpaquette:
Addresses a change to the db_connect function to return a sessionMaker object instead of a session like the PDS_DBquery object was expecting.

Fixes #372

### DIFF
--- a/pds_pipelines/config.py
+++ b/pds_pipelines/config.py
@@ -1,20 +1,20 @@
 import os
 # Database credentials
-credentials = {'upc_test': {'user': 'mgr',
-                        'pass': '1234',
+credentials = {'upc_test': {'user': 'postgres',
+                        'pass': '',
                         'host': 'localhost',
-                        'port': '9010',
+                        'port': '5432',
                         'db': 'upc_test'},
-               'di_test': {'user': 'mgr',
-                            'pass': '1234',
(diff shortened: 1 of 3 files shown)


kaitlyndlee:
Could you update the db_connect() function in db.py to explain what the return values session and engine are? I guess it is a little misleading to call it a session when it actually returns a session_maker

acpaquette:
@kaitlyndlee Addressed your comment regarding function use/returns being unclear

PR #400, opened 2020-04-20
Search Term Mapping

acpaquette:
Adds an optional read from the upc section of a recipe file to remap the search_term_attributes to other keywords to search for.

### DIFF
--- a/pds_pipelines/upc_process.py
+++ b/pds_pipelines/upc_process.py
@@ -444,7 +444,7 @@ def generate_processes(inputfile, recipe_string, logger):
     no_extension_inputfile = os.path.splitext(inputfile)[0]
     cam_info_file = os.path.splitext(inputfile)[0] + '.caminfo.pvl'
 
-    template = jinja2.Template(recipe_str)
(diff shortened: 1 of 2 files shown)


pds_pipelines/upc_process.py
```diff
@@ -544,7 +543,16 @@ def main(user_args):
...
         context['inputfile'] = inputfile
 
-        processes, infile, caminfoOUT, workarea_pwd = generate_processes(inputfile, archive, logger)
+        recipe_file = recipe_base + "/" + archive + '.json'
+        with open(recipe_file) as fp:
+            upc_json = json.load(fp)['upc']
+            recipe_string = json.dumps(upc_json['recipe'])
+            try:
+                search_term_mapping = upc_json['search_term_mapping']
+            except:
```
AustinSanders, line 552: Is there a more explicit exception type we can catch here?

acpaquette: :+1: KeyError. I'll get this updated

PR #407, opened 2020-04-22
Initial GDAL wrappers

AustinSanders:
Initial wrappers for GDAL functions.

I prefer function definitions to be func(src, dest), but the api calls are all func(dest, src), so I kept that convention.

### DIFF
--- a/pds_pipelines/__init__.py
+++ b/pds_pipelines/__init__.py
@@ -11,6 +11,38 @@ except Exception as e:
 try:
     import gdal
     available_modules['gdal'] = gdal
+    # Python GDAL doesn't use exceptions by default.
+    gdal.UseExceptions()
 except Exception as e:
     warnings.warn('Unable to add gdal to the available modules. ' +
                  f'Failed with the following {e}.')
+
+try:
+    import ogr
+    available_modules['ogr'] = ogr
+except Exception as e:
+    warnings.warn('Unable to add ogr to the available modules. ' +
+                 f'Failed with the following {e}.')
+
+
+def gdal_translate(dest, src, **kwargs):
+    options = gdal.TranslateOptions(**kwargs)
+    return gdal.Translate(dest, src, options)
+
+
+def gdal_polygonize(band, mask, output_name, **kwargs):
+    driver = ogr.GetDriverByName("ESRI Shapefile")
+    output_datasource = driver.CreateDataSource(output_name + ".shp")
+    out_layer = output_datasource.CreateLayer(output_name, srs=None)
+    field = ogr.FieldDefn('MYFLD', ogr.OFTInteger)
+    out_layer.CreateField(field)
+    return gdal.Polygonize(band, mask, out_layer, -1, [], **kwargs)
+
+
+def ogr2ogr(dest, src, **kwargs):
+    srcDS = gdal.OpenEx(src)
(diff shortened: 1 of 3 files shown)


pds_pipelines/__init__.py
```diff
@@ -11,6 +11,38 @@
...
+    import ogr
+    available_modules['ogr'] = ogr
+except Exception as e:
+    warnings.warn('Unable to add ogr to the available modules. ' +
+                 f'Failed with the following {e}.')
+
+
+def gdal_translate(dest, src, **kwargs):
+    options = gdal.TranslateOptions(**kwargs)
+    return gdal.Translate(dest, src, options)
```
acpaquette, line 30: We'll need to update how the process function accesses these functions since they won't be part of the available_modules dictionary. That or we will need to add them to the dictionary in some fashion

pds_pipelines/__init__.py
```diff
@@ -11,6 +11,38 @@
...
+
+try:
+    import ogr
+    available_modules['ogr'] = ogr
+except Exception as e:
+    warnings.warn('Unable to add ogr to the available modules. ' +
+                 f'Failed with the following {e}.')
+
+
+def gdal_translate(dest, src, **kwargs):
```
dpmayerUSGS, line 28: I can get this to work as expected:
```
def gdal_translate(dest, src, **kwargs):
    topts = gdal.TranslateOptions([], **kwargs)
    return gdal.Translate(dest, src, options=topts)
```

recipe/new/mro_hirise_rdr.json
```diff
@@ -1,34 +1,31 @@
...
+                "scaleParams": [[1,65535,1,1]],
+                "noData": "0",
+                "outputType": "gdal.GDT_Byte",
+                "xRes": "25",
+                "yRes": "25",
+                "src" : "{{no_extension_inputfile}}.cub",
+                "dest": "{{no_extension_inputfile}}.tif"
             },
             "gdal_polygonize": {
-                "8",
```
AustinSanders, line 20: It wouldn't be hard to support the -8.  It looks like this can be passed as an option "8connected=8," but there doesn't seem to be an associated keyword in the [documentation](https://github.com/OSGeo/gdal/blob/2fde97fd57bbd892781f89323499528c5821dea4/gdal/swig/python/scripts/gdal_polygonize.py)

dpmayerUSGS: I think it's okay to omit it from this initial implementation. It shouldn't actually matter because we are expecting a single, hole-free polygon, but we can consider implementing this in the future.

recipe/new/mro_hirise_rdr.json
```diff
@@ -8,27 +8,24 @@
...
             },
             "ogr2ogr": {
-                 "f": "GeoJSON",
-                 "t_srs": ""+proj=longlat +R=3396190 +no_defs"",
-                 "{{no_extension_inputfile}}_footprint.json",
-                 "{{no_extension_inputfile}}_footprint.shp"
-            },
+                 "format": "GeoJSON",
+                 "dstSRS": "+proj=longlat +R=3396190 +no_defs",
+                 "dest": "{{no_extension_inputfile}}_footprint.json",
```
acpaquette, line 26: There is a new variable in the ninja template for a footprint file. This should probably utilize that

PR #410, opened 2020-04-23
Removed extra try/except in upc_process

kaitlyndlee:
If an isis id is not found or pyisis is not imported, than we set the isisid to None. I don't know if we need to catch an Attribute Error. I think I originally wrote that test and may have chosen an error that wasn't Key or Process just to get isis id set to None.

### DIFF
--- a/pds_pipelines/tests/test_upc.py
+++ b/pds_pipelines/tests/test_upc.py
@@ -188,7 +188,7 @@ def test_datafiles_no_label(mocked_pds_id, mocked_isis_id, session, session_make
 def test_datafiles_no_isisid(mocked_pds_id, session, session_maker, pds_label):
     # Since we mock getsn above, make it throw an exception here so we can test
     # when there is no ISIS ID.
-    with patch('pysis.isis.getsn', side_effect=Exception('ProcessError')):
+    with patch('pysis.isis.getsn', side_effect=ProcessError(1, ['getsn'], '', '')):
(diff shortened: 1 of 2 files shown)


pds_pipelines/tests/test_upc.py
```diff
@@ -188,7 +188,7 @@ def test_datafiles_no_label(mocked_pds_id, mocked_isis_id, session, session_make
 def test_datafiles_no_isisid(mocked_pds_id, session, session_maker, pds_label):
     # Since we mock getsn above, make it throw an exception here so we can test
     # when there is no ISIS ID.
-    with patch('pysis.isis.getsn', side_effect=Exception('AttributeError')):
+    with patch('pysis.isis.getsn', side_effect=Exception('ProcessError')):
```
acpaquette, line 191: This needs to change from an Exception constructor to a ProcessError constructor.

So `side_effect=Exception()` should be `side_effect=ProcessError()`

There should be a few things you need to pass to the constructor as well.

pds_pipelines/upc_process.py
```diff
@@ -262,7 +258,10 @@ def create_datafiles_record(label, edr_source, input_cube, session_maker):
     # Attemp to get the ISIS serial from the cube
     try:
         isis_id = getISISid(input_cube)
-    except:
+        
+    except(ProcessError, KeyError):
+        # If either isis was not imported or a serial number could not be
+        # generated from the infile set the serial number to an empty string
         isis_id = None
```
acpaquette, line 265: I know this is only used once in all of PDS pipelines but I think it serves a better place in the getISISid function since  we would only need to actually handle the error in that one place.

Issue #284, opened 2019-08-26
UPC_process.py: Certain galileo_ssi_edr ProductIds Converted to Float By Mistake


dpmayerUSGS:
**Describe the bug**
`galileo_ssi_edr` products have ProductIDs that are 7 characters long and take the format `[0-9][0-9][A-Z][0-9][0-9][0-9]`. If the letter in the productID happens to be `E`, the product ID gets interpreted by the PVL library as a number, instead of a string. 

**To Reproduce**
Queue up a volume of Galileo data that includes observations of Europa and run UPC
```
python UPCqueueing.py --archive galileo_ssi_edr --volume go_0022
python HPCjobber.py --process upc
```
The `productid` field in the `datafiles` table will contain values for images of Europa that look like the floating point number `250.0` instead of as the string `25E0010`. Also, the JSON representation of the label in the `json_keywords` table will represent the productid as the floating point number `250.0` instead of as the string `25E0010`.

**Expected behavior**
ProductId values should always be interpreted and represented as strings wherever they appear in the UPC database.

**Additional context**
This is fundamentally an edge case in the PVL library rather than a bug in `UPC_process.py` per se, but it may not be practical to effect a change in the PVL library, so I'm opening the issue here. The relevant part of the PVL library that parses numbers is https://github.com/planetarypy/pvl/blob/master/pvl/_numbers.py

As far as I can tell, `galileo_ssi_edr` products are currently the only ones with ProductIds that could trigger this edge case in the PVL library.

**Proposed Solution**
Ideally, there would be a way for a user to force the PVL decoder to read a specific keyword as a string in order to avoid treating the value as a number when it shouldn't. But in the absence of that functionality in PVL there could be a subprocess call in `UPC_process.py` to the ISIS program `getkey` to extract the ProductId from the product label, rather than using the PVL library.

PR #411, opened 2020-04-23
Record Delete and Minor Fixes

acpaquette:
Adds a script to delete records from the di and upc databases based on the productid from the UPC database.

This also addresses an issue where the productid was not being populated as the file trying to be read did not exist.

### DIFF
--- a/pds_pipelines/models/upc_models.py
+++ b/pds_pipelines/models/upc_models.py
@@ -2,7 +2,7 @@ import sqlalchemy
 
 from sqlalchemy.ext.declarative import declarative_base, declared_attr
 from sqlalchemy_utils import database_exists, create_database
-from sqlalchemy import (Column, Integer, Float, Time, String, Boolean, 
+from sqlalchemy import (Column, Integer, Float, Time, String, Boolean,
                         ForeignKey, CHAR, DateTime)
 from sqlalchemy.dialects.postgresql import JSONB
 from sqlalchemy.orm import relationship
@@ -60,8 +60,8 @@ class DataFiles(BaseMixin, Base):
     detached_label = Column(String(1024))
     instrumentid = Column(Integer, ForeignKey("instruments.instrumentid"))
     targetid = Column(Integer, ForeignKey("targets.targetid"))
-    search_terms = relationship('SearchTerms', backref="DataFiles", uselist=False)
-    json_keyword = relationship('JsonKeywords', backref="DataFiles", uselist=False)
+    search_terms = relationship('SearchTerms', backref="DataFiles", uselist=True, cascade="all, delete-orphan")
+    json_keyword = relationship('JsonKeywords', backref="DataFiles", uselist=True, cascade="all, delete-orphan")
     level = Column(CHAR(1))
 
 
--- a/pds_pipelines/models/upc_models.py
+++ b/pds_pipelines/models/upc_models.py
@@ -60,8 +60,8 @@ class DataFiles(BaseMixin, Base):
     detached_label = Column(String(1024))
     instrumentid = Column(Integer, ForeignKey("instruments.instrumentid"))
     targetid = Column(Integer, ForeignKey("targets.targetid"))
-    search_terms = relationship('SearchTerms', backref="DataFiles", uselist=True, cascade="all, delete-orphan")
-    json_keyword = relationship('JsonKeywords', backref="DataFiles", uselist=True, cascade="all, delete-orphan")
+    search_terms = relationship('SearchTerms', backref="DataFiles", uselist=True, cascade="save-update, merge, delete, delete-orphan")
+    json_keyword = relationship('JsonKeywords', backref="DataFiles", uselist=True, cascade="save-update, merge, delete, delete-orphan")
     level = Column(CHAR(1))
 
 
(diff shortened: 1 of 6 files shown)


pds_pipelines/upc_process.py
```diff
@@ -261,15 +267,15 @@ def create_datafiles_record(label, edr_source, input_cube, session_maker):
 
     # Attemp to get the ISIS serial from the cube
     try:
-        isis_id = getISISid(input_cube)
+        isis_id = getISISid(input_cube + '.cub')
```
acpaquette, line 270: This needs to change as expecting some step in the recipe to produce a just '.cub' extension is extremely fragile

acpaquette:
I'm expecting tests for fail on this initially and will update once the failures come through

pds_pipelines/upc_process.py
```diff
@@ -29,11 +29,12 @@
 
 
 def getPDSid(infile):
-    """ Use ISIS to get the PDS Product ID of a cube.
+    """
+    Use ISIS to get the PDS Product ID of a cube.
```
dpmayerUSGS, line 33: Docstring should be updated. This function does not use `isis.getkey()` to extract the PDS product ID, rather, it uses PVL. The use of the PVL library to extract the product ID may have reintroduced the bug in #284 , but I'm okay with re-opening that after this PR is merged.

pds_pipelines/upc_process.py
```diff
@@ -47,7 +48,10 @@ def getPDSid(infile):
         The PDS Product ID.
     """
     upc_keywords = UPCkeywords(infile)
-    prod_id = upc_keywords.getKeyword('productid')
+    for key in ['productid', 'product_id', 'PRODUCTID', 'PRODUCT_ID']:
```
dpmayerUSGS, line 51: I think the ALL CAPS items can be removed because `getKeyword()` forces averything to lowercase:
https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/777f7a5bc96bbb21f5a4df0bd6bcb65baef80331/pds_pipelines/upc_keywords.py#L50

pds_pipelines/upc_process.py
acpaquette: @dpmayerUSGS This issue seems to only crop up in isis cubes, could we use the original label to extract the product ID? Or is there some other edge case around that?

pds_pipelines/models/upc_models.py
```diff
@@ -60,8 +60,8 @@ class DataFiles(BaseMixin, Base):
     detached_label = Column(String(1024))
     instrumentid = Column(Integer, ForeignKey("instruments.instrumentid"))
     targetid = Column(Integer, ForeignKey("targets.targetid"))
-    search_terms = relationship('SearchTerms', backref="DataFiles", uselist=False)
-    json_keyword = relationship('JsonKeywords', backref="DataFiles", uselist=False)
+    search_terms = relationship('SearchTerms', backref="DataFiles", uselist=True, cascade="save-update, merge, delete, delete-orphan")
+    json_keyword = relationship('JsonKeywords', backref="DataFiles", uselist=True, cascade="save-update, merge, delete, delete-orphan")
```
acpaquette, line 64: This sort of change requires a database delete and reup/table drop of the changed tables. Just to be clear that this delete script won't work on the current database.

pds_pipelines/upc_process.py
dpmayerUSGS: As per offline conversation, the short answer is "no." Unfortunately, the value that is understood by humans to be the PDS ProductID can appear under many (potentially arbitrarily many) keywords in the original PDS label, and varies by dataset.

tools/delete_record.py
```diff
@@ -0,0 +1,78 @@
...
+from pds_pipelines.models.upc_models import DataFiles
+from pds_pipelines.models.pds_models import Files
+from pds_pipelines.config import credentials, upc_db, pds_db
+
+def parse_args():
+    parser = argparse.ArgumentParser(description='Database Record Delete')
+    parser.add_argument('--pdsid', '-r', dest='pdsid', help="PDS id to delete from " +
+                        "the selected databases", type=str)
+    parser.add_argument('-di', dest="di", help="Flag defining if the record should " +
+                        "be deleted from the di database", action='store_true')
```
dpmayerUSGS, line 15: "di" --> "DI" in the help text

tools/delete_record.py
```diff
@@ -0,0 +1,78 @@
...
+    parser.add_argument('-upc', dest="upc", help="Flag defining if the record should " +
+                        "be deleted from the upc database", action='store_true')
```
dpmayerUSGS, line 17: "upc" --> "UPC" in the help text

pds_pipelines/models/upc_models.py
dpmayerUSGS: Understood. Shouldn't be an issue for UPC because we're going to be building that db up from scratch. We have several options for dealing with the DI db.

tools/delete_record.py
```diff
@@ -0,0 +1,78 @@
...
+        pds_session = pds_session_maker()
+
+        query_res = pds_session.query(Files).filter(
+                    Files.filename.contains(pds_id))
+        num_pds_queries = len(list(query_res))
+
+        while(True):
+            print(f'You will be deleteing {num_pds_queries} from the di ' +
+                  f"database {credentials[pds_db]['db']}")
+            user_answer = input('Are you sure?[Y/N]:')
```
AustinSanders, line 39: It would be more user-friendly to .upper() the user_answer here

PR #414, opened 2020-04-27
Minor changes to UPC process to support insertion into DB + small recipe / doc changes

AustinSanders:


acpaquette:
@AustinSanders You should do a rebase on this to get the old commits out of the PR

Issue #419, opened 2020-04-29
Move pds_pipelines available _modules into subpackage


dpmayerUSGS:
**Description**
Thin wrappers for various ISIS and GDAL/OGR functions are currently stored inside `pds_pipelines/__init__.py`. This enables processes like UPC to be more flexible in which programs can be run on input data, and how those programs can be parameterized. The approach of building an `available_modules` dictionary inside `__init__.py` isn't a bad idea in theory, but because `pysis` actually changes the system `$PATH` when it's imported, it causes problems for ways in which other parts of PDS-Pipelines are used in practice.

For example, POW/MAP2 jobs are read from a Redis queue by `service_job_manager.py`, which writes an sbatch script containing a call to either `pow_process.py` or `map_process.py` and the associated jobkey,  and submits the script to SLURM. However, because `serve_job_manager.py` (naturally) imports parts of `pds_pipelines`, it causes `pysis` to be imported, even though the service job manager doesn't use it. `pysis` prepends the path to ISIS executables to $PATH, which SLURM inherits. The `python` interpreter from ISIS is now at the front of $PATH, instead of the one associated with the conda environment. `{pow,map}_process.py` inside the sbatch script then fails with `ModuleNotFoundError` when the Python script attempts to import a module that the "ISIS python" doesn't know about.

**Steps to Reproduce**
1. Submit a POW or MAP2 job
2. run `service_job_manager.py -k <jobkey>` to submit the job to SLURM
3. Locate the `ModuleNotFound` error in SLURM log and note that the SLURM job exited with "FAILED" status

By contrast, after the SLURM job fails in the example above, the sbatch script will still be available in the work area and can be submitted from the nominal environment for PDS-Pipelines with `sbatch <jobkey>.sbatch` and it will run as expected because SLURM will inherit the correct shell environment, including the correct $PATH.

**Proposed Solution**
Take the current contents of `pds_pipelines/__init__.py` and move them into a subpackage, perhaps `pds_pipelines/available_modules.py`. 

**Definition of Done**
A solution should only be considered acceptable if 
- [ ] `upc_process.py` continues to function normally (that is, can access the functions currently in `__init__.py`)
- [ ] The `service_job_manager.py` --> SLURM workflow described above runs without error

PR #423, opened 2020-05-04
Added recursive option to getkey call in upc_process, force getkey to…

### LINKED ISSUE #421: upc_process.py: `getkey` doesn't find ProductId
**Description**
`upc_process.py` uses the ISIS `getkey` command to extract the PDS ProductId from a given cube. 
As currently written in `upc_process.py`, `getkey` won't find ProductId unless it is at the root level of the cube label (which would be unusual). The ProductId keyword can be stored in a different group inside the cube depending on whether the cube is based on a Level-1 or Level-2 PDS product. 

**Proposed Solution**
Add the `recursive=TRUE` option to the `getkey` command.

https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/180d9189312c96cb13a8a852dfd2b9b4bb9eee2d/pds_pipelines/upc_process.py#L56

### LINKED ISSUE #422: upc_process.py: Incorrect file being passed to getPDSid()
**Description**
The `getPDSid()` function is supposed to take as input any file that the ISIS `getkey` command can parse. In practice, this input file is an ISIS cube, and is named like the basename of the original PDS file + ".cub". 

`getPDSid()` is called in 2 places in `upc_process.py`:
https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/180d9189312c96cb13a8a852dfd2b9b4bb9eee2d/pds_pipelines/upc_process.py#L288

https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/180d9189312c96cb13a8a852dfd2b9b4bb9eee2d/pds_pipelines/upc_process.py#L364

However, the variable `input_cube` is not actually an ISIS cube, but just a string containing the value of `no_extension_inputfile`, which itself if coming from [`generate_processes()`](https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/180d9189312c96cb13a8a852dfd2b9b4bb9eee2d/pds_pipelines/upc_process.py#L471). 

This causes `getPDSid()` to fail to extract the PDS ProductId (thereby leaving the corresponding fields blank in the search_terms and datafiles tables) because it is passed a file that does not exist.

**Proposed Solution**
In current practice, the file that we want to pass to `getPDSid()` can reasonably be expected to be named like "`no_extension_inputfile` + .cub". It might be worth creating another jinja template that specifies a file named that way. Then that template can be used in the individual recipes to clearly indicate which processing step creates the file that should be used to extract the PDS ProductID.

For example, inside `generate_processes()`, we could define `isis_cube = no_extension_inputfile + ".cub"` and then add an item to the jinja template renderer like `isis_cube = isis_cube`. In place of `no_extension_inputfile`, return `isis_cube`. (`no_extension_inputfile` doesn't actually get used for anything else inside `upc_process.py`; it's just convenient for use in naming files in recipes.)

This proposed solution would *require* editing UPC recipes for *all* supported datasets to use the new "isis_cube" template.

Alternatively, we could just stick `+ ".cub"` on the end of `input_cube` in the 2 places where `getPDSid()` is used.

dpmayerUSGS:
… look for file with .cub extension, added cubeatt step to THEMIS VIS EDR UPC recipe to make sure there's a file with name getkey is expecting.

Fixes #421 
Fixes #422

### DIFF
--- a/pds_pipelines/tests/test_upc.py
+++ b/pds_pipelines/tests/test_upc.py
@@ -21,7 +21,7 @@ def getsn(from_):
 pysis.isis.getsn = getsn
 
 # Stub in the getkey function for testing
-def getkey(from_, objname=None, grp=None, keyword=None, keyindex=None, upper=False, recursive=False):
+def getkey(from_, objname=None, grp=None, keyword=None, keyindex=None, upper=False, recursive=True):
     return b'PRODUCTID'
 pysis.isis.getkey = getkey
 
--- a/pds_pipelines/tests/test_upc.py
+++ b/pds_pipelines/tests/test_upc.py
@@ -175,6 +175,7 @@ def test_datafiles_insert(mocked_pds_id, mocked_isis_id, session, session_maker,
     input_cube = '/Path/to/my/cube.cub'
     upc_id = create_datafiles_record(pds_label, '/Path/to/label/location/label.lbl', '/Path/to/my/cube', session_maker)
     mocked_isis_id.assert_called_with(input_cube)
+    input_cube = '/Path/to/my/cube'
     mocked_pds_id.assert_called_with(input_cube)
 
     resp = session.query(models.DataFiles).filter(models.DataFiles.isisid=='ISISSERIAL').first()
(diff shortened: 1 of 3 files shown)


dpmayerUSGS:
I can't figure out how to update `test_datafiles_insert()` to work with the change I made to `getPDSid()` to fix #419

PR #424, opened 2020-05-05
Move processing modules for UPC into own subpackage

### LINKED ISSUE #419: Move pds_pipelines available _modules into subpackage

dpmayerUSGS:
Fixes #419 
 - Moved modules and funcs for UPC processing out of `__init__.py` and into own subpackage
 - Updated upc_process.py to correctly import/access available_modules. 

As per the"Definition of Done" in #419 , I tested by making sure that I could queue and process (using SLURM) a HiRISE RDR, and verified the expected output was written to the UPC database. I similarly tested UPC processing on a THEMIS VIS EDR to make sure that Level-1 support didn't break. I then invoked `service_job_manager.py` on a specific MAP2 jobkey and successfully processed a MAP2 job, up to and including final compression/copying to the download area.

Also:
 - Corrected error in path to HiRISE archive in `PDSinfo.json`
 - Updated UPC recipe for HiRISE RDRs
   - Now dumps ISIS label + original label to an actual PVL file using `caminfo`, which eliminates conflicts with `upc_process.py`'s expectations for what certain files should be named
   - Corrected where "footprint_file" should appear in the recipe


dpmayerUSGS:
I expect tests to fail, as per my comment in #423

recipe/new/mro_hirise_rdr.json
```diff
@@ -18,12 +26,12 @@
...
             "gdal_polygonize": {
                 "input_file": "{{no_extension_inputfile}}.tif",
-                "output_name": "{{footprint_file}}"
+                "output_name": "{{no_extension_inputfile}}_footprint.shp"
             },
             "ogr2ogr": {
                  "format": "GeoJSON",
                  "dstSRS": "+proj=longlat +R=3396190 +no_defs",
-                 "dest": "{{no_extension_inputfile}}_footprint.json",
+                 "dest": "{{footprint_file}}",
```
USGSsakins, line 34: Why did we drop the file extension here?

dpmayerUSGS: @USGSsakins This is actually a correction to bug in the recipe that I discovered while I was testing #419 .
`{{footprint_file}}` is meant to represent the name of the JSON file that `upc_process.py` reads in cases where we don't get the footprint out of the PVL. The [definition](https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/e56cd03d3af5478e89dee99371c6a75227875627/pds_pipelines/upc_process.py#L473) includes the file extension, so it was a bug to have a file extension listed in the recipe in the first place.

I think it's justifiable to include the file extension in the definition of `{{footprint_file}}` because the function that will actually consume it is expecting JSON.

Furthermore, `{{footprint_file}}` had been specified as the output of the `gdal_polygonize` step, but it's supposed to be the output of the `ogr2ogr` step that follows `gdal_polygonize`. The `ogr2ogr` step is supposed to take the output of `gdal_polygonize` as input, so I corrected that as well.

USGSsakins: Okay, I just wanted to be sure.  Sounds good.

PR #447, opened 2020-07-27
Skip filter logic if only one candidate key is found

### LINKED ISSUE #443: service_job_manager.py: getCleanName() filtering by upc_reqs is unreliable for archives detached labels
**Description**
The `getCleanName()` function attempts to get the "clean name" of the archive that images in a particular POW job come from in order to determine which recipe file needs to be loaded. It does this by attempting to match the front of the file path to paths listed in `PDSinfo.json,` and then filtering these matches by the `upc_reqs` items for each of the candidate archives. However, `upc_reqs` are really only for identifying files for the purposes of Ingest and UPC processing. For archives with detached labels, PILOT usually returns a list of `*.img` files, rather than their labels. `getCleanName()` fails to find any matches for candidate archives because `upc_reqs` only include the `.lbl` or `.LBL` in the case of archives with detached labels. When the function fails to find any matches, `service_job_manager.py` silently exits without generating a `*.sbatch` file and launching a processing job.

**Steps to Reproduce**
Submit a POW job containing 1 or more images from an archive with detached labels, such as Galileo SSI. Note in `Service.log` that the job simply stops after logging the message "INFO, Building Recipe". 

**Proposed Solution**
The [current logic](https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/06ba9e3bf9c07f01d24914041cc8b159b741be1b/pds_pipelines/service_job_manager.py#L66-L72) attempts to filter by `upc_reqs` even if there is only 1 candidate returned by the initial search. Simply adding a condition to only perform the filtering step if `candidates > 1` should be enough to avoid this issue.
In practice, filtering by `upc_reqs` shouldn't be necessary for datasets currently supported by POW. An alternative could be to remove the filtering step altogether, but this seems to invite a future problem where there may be a genuine ambiguity when trying to identify the dataset based on its path alone.

**Additional Context** 
The `getCleanName()` function is admittedly a hacky workaround for dealing with the fact that job XML does not explicitly identify the name of the dataset that the images come from. Changing that is not something entirely within the scope of PDS-Pipelines itself.

AustinSanders:
Since this is a process-breaking issue, I decided to take the low hanging fruit and get this pushed out quickly.

It's worth looking into alternate fixes -- probably something that doesn't use upc_reqs.  upc_reqs isn't really supposed to be a mechanism for uniquely identifying archives, but it initially seemed like it was sufficient for that purpose.  The existence of this issue seems to suggest otherwise.

On the other hand, the upc_reqs based filtering can be considered a "last ditch effort" in that it provides an additional layer of information that can be used to uniquely identify an archive that is otherwise not uniquely identifiable.

Closes #443

### DIFF
--- a/pds_pipelines/service_job_manager.py
+++ b/pds_pipelines/service_job_manager.py
@@ -63,13 +63,14 @@ class jobXML(object):
 
         # try to filter list based on upc_reqs.  If upc_reqs isn't specified, just skip the filtering step
         # ps can't use 'filter' because not all elements have upc_reqs, so they may raise exceptions.
-        for item in candidates:
-            try:
-                if not all(x in file_name for x in self.pds_info[item]['upc_reqs']):
-                    candidates.remove(item)
-            except KeyError:
-                # Intentionally left blank.  Unspecified upc_reqs is valid -- there's just nothing to do for those elements
-                pass
+        if len(candidates) > 1:
+            for item in candidates:
+                try:
+                    if not all(x in file_name for x in self.pds_info[item]['upc_reqs']):
+                        candidates.remove(item)
+                except KeyError:
+                    # Intentionally left blank.  Unspecified upc_reqs is valid -- there's just nothing to do for those elements
+                    pass
 
         # If multiple candidates still exist, then it is not possible to uniquely identify the clean name
         if len(candidates) > 1:


dpmayerUSGS:
+1 for opening a new issue to address the root problem associated with upc_reqs. The PR itself looks good to me.

PR #456, opened 2020-07-28
Less Greedy File Removal

### LINKED ISSUE #436: upc_process.py: Make glob pattern for removing files less greedy
The glob pattern for removing files from the workarea directory is too greedy. This has the undesirable effect of deleting *all* files in the working directory, but some of those files may be part of UPC processing jobs running in parallel. 

The glob pattern should be made more specific to only remove files from the working directory that match the basename of `inputfile`. 

https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/06ba9e3bf9c07f01d24914041cc8b159b741be1b/pds_pipelines/upc_process.py#L616

acpaquette:
Only remove files that contain the name of the original file being passed in when no persist flag is present.

Closes #436


dpmayerUSGS:
@acpaquette please re-sync and then I think this will be good to squash+merge

Issue #465, opened 2020-07-30
Update POW, browse/thumbnail recipes to use jinja templates


dpmayerUSGS:
This issue anticipates the implementation of features described in #441, #442, and #458 .
POW and browse/thumbnail recipes will need to be reformatted in order to be compatible with the new jinja template support in the corresponding processing scripts. Specifically, the filename arguments in existing POW recipes need to be modified to use jinja template tokens. This issue lists a subset of datasets for which existing recipes need to be converted to work with the jinja templates.

**Definition of done**
All the POW and browse/thumbnail recipes for datasets in the checklist below are modified to use the jinja tokens as appropriate, following the example of the existing UPC recipes. POW recipes for mro_ctx and mgs_moc will also have dataset-specific if/then tests to determine if certain ISIS applications should be called as part of the recipe.
 - [ ] mro_ctx
  - if `label['IsisCube']['Instrument']['SpatialSumming'] = 1` based on label `{{input_file_no_extension}}.cub`, run `ctxevenodd`
 - [x] lro_lroc_edr - Austin
 - [x] galileo_ssi_edr
 - [x] cassini_iss
 - [x] mgs_moc
  - if `label['IsisCube']['Instrument']['CrosstrackSumming'] = 1` based on label `{{input_file_no_extension}}.cub`, run `mocevenodd`
 - [x] mo_themis_vis_edr
  - browse/thumbnail only for now because not POW-eligible
 - [x] mo_themis_ir_edr
 - [x] messenger

**Additional notes**
This is issue should only be closed once all items in the list above are checked off, but PRs can include changes to only a single recipe. PRs that modify recipes as described above should mention this issue.

dpmayerUSGS:
Partially addressed in #483

Issue #469, opened 2020-08-03
Queueing / copying functionality is a good candidate for modularizing / converting to functions.


AustinSanders:
Queueing and copying functionality in [upc, browse, thumbnail]_queueing.py is almost a straight copy / paste.

Converting this to a function would increase maintainability and allow us to write tests.

dpmayerUSGS:
Closed by #527

PR #470, opened 2020-08-04
Derived processing refactor

### LINKED ISSUE #441: browse_process.py: Make image processing loop more generic
The current browse processing script uses a large and unwieldy if/then/else ladder to parameterize image processing steps from a recipe by matching the name of the program being called in the step. It would be better if this logic could be replaced with something that can consume and execute browse recipes in a more generic way.

The preferred solution is to replace the processing logic with a system more like the one in `upc_process.py`, where there is a [function to parse an input recipe and generate a parameterized list of processing steps](https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/06ba9e3bf9c07f01d24914041cc8b159b741be1b/pds_pipelines/upc_process.py#L464) using jinja templates, and then a [function to consume and execute those steps](https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/06ba9e3bf9c07f01d24914041cc8b159b741be1b/pds_pipelines/upc_process.py#L484).

The proposed solution will require changes to existing browse recipes in order to set up find/replace tokens for jinja and to alter the names of the programs to be called in the recipe, but these should be tracked in a separate issue.

### LINKED ISSUE #442: thumbnail_process.py: Make image processing loop more generic
The current thumbnail processing script uses a large and unwieldy if/then/else ladder to parameterize image processing steps from a recipe by matching the name of the program being called in the step. It would be better if this logic could be replaced with something that can consume and execute browse recipes in a more generic way.

The preferred solution is to replace the processing logic with a system more like the one in `upc_process.py`, where there is a [function to parse an input recipe and generate a parameterized list of processing steps](https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/06ba9e3bf9c07f01d24914041cc8b159b741be1b/pds_pipelines/upc_process.py#L464) using jinja templates, and then a [function to consume and execute those steps](https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/06ba9e3bf9c07f01d24914041cc8b159b741be1b/pds_pipelines/upc_process.py#L484).

The proposed solution will require changes to existing thumbnail recipes in order to set up find/replace tokens for jinja and to alter the names of the programs to be called in the recipe, but these should be tracked in a separate issue.

### LINKED ISSUE #453: Merge thumbnail & browse into single processing script
`browse_process.py` and `thumbnail_process.py` are nearly identical. It would make sense to merge them into a single script that can consume the contents of *either* the `Browse_ReadyQueue` or `Thumbnail_ReadyQueue` and execute browse or thumbnail recipes as appropriate.

The central goal of #442 and #441 to make the image processing part of the thumbnail and browse scripts more generic remains, and is in fact central to enabling the scripts to be combined.

*Edited to add*: This should include modifying the processing script to act on a single file from a given queue (like upc_process.py) to that many files can be processed in parallel using SLURM.

### LINKED ISSUE #459: {browse,thumbnail}_queueing.py: Add logic to copy input files if needed
As part of the browse/thumbnail processing refactor to be more like UPC processing, it makes sense to modify the corresponding queueing scripts to be more like upc_queueing.py in the following manner:

- As part of queueing, copy input files from the archive base into the workarea, recreating the directory structure as needed. 
- Take advantage of the possibility the upc_process.py has left a copy of the input files in the workarea already by checking to see if the queued files already exists there, and skip the copying step.

### LINKED ISSUE #460: browse/thumbnail process: Update where image URLs are inserted in UPC database
Historically, the browse/thumbnail process scripts inserted records into the "meta_strings" table of the UPC database containing the URLs to the browse/thumbnail images associated with a given upcid. This table was removed as part of the UPC database refactor, so the browse/thumbnail processing scripts will need to be updated to insert the URLs in a different place.

It would make sense to append key:value pairs for the browse/thumbnail URLs inside the `jsonkeywords` field in the `json_keywords` table of the UPC database. An application, such as PILOT, that queries the UPC database would parse the contents of `json_keywords` for a given `upcid`and then look up the URLs.

I propose using `browse` and `thumbnail` as the keys. The URLs would be created

For example, a `json_keywords` record containing the following
```
{
    "caminfo": {
        "polygon": {
            "mapping": {
                "targetname": "IO",
                "polarradius": [
                    1815700,
                    "meters"
                ],
                   << etc. >>
            }
        }
    }
}
```

Would be updated to contain the browse/thumbnail URLs like this:
```
{
    "caminfo": {
        "polygon": {
            "mapping": {
                "targetname": "MARS",
                "polarradius": [
                    3376200,
                    "meters"
                ],
                   << etc. >>
            }
        }
    },
   "browse": {
      "https://upcimages.wr.usgs.gov/mars_reconnaissance_orbiter/ctx/mrox_1391/data/G14_023781_0920_XN_88S062W.browse.jpg"},
   "thumbnail": {
      "https://upcimages.wr.usgs.gov/mars_reconnaissance_orbiter/ctx/mrox_1391/data/G14_023781_0920_XN_88S062W.thumbnail.jpg"},
}
```

If #457 is implemented, then it will be easy to query the `json_keywords` table in the UPC database from the browse/thumbnail processing script and find the entry associated with a product based on its (unique) `upcid`.

### LINKED ISSUE #468: "Search" argument is unused in browse/thumbnail queueing
Users can specify a search string in browse / thumbnail queueing, but the option is unused.  In other words, there is no logic implemented for the search.

Search functionality should be the same as in upc_queueing.

### LINKED ISSUE #471: derived_process.py: Replace $browse_server with variable
Historically, the browse/thumbnail processing script would write paths to the browse/thumbnail JPEGs into the UPC database that contained variables such as `$browse_server` that were meant to be interpreted by PILOT.

Rather than expecting a front end application to perform some find/replace action, the path that is written into the database should be a fully-qualified URL. To do this, a new variable, `derived_base` should be defined in `config.py` whose value is "https://upcimages.wr.usgs.gov" by default. The processing script should be modified as needed to join `derived_base` and any path prefix such that a fully-qualified URL gets written to the database. 

A real example would look like "https://upcimages.wr.usgs.gov/mars_reconnaissance_orbiter/ctx/mrox_1391/data/G14_023781_0920_XN_88S062W.thumbnail.jpg"

### LINKED ISSUE #472: Combine thumbnail / browse queueing
Thumbnail and browse queueing are virtually identical.  This results in duplication not only in code, but also in processing.  Ultimately, the only difference in processing is the size of output images, but we run the full "derived" recipe once per process type.

We could fairly easily combine these into a single script, queue, and recipe by calling gdal_translate twice within the derived recipe.

AustinSanders:
This PR combines thumbnail / browse processing scripts into a unified "derived_process.py."  As part of this PR, it was necessary to update the queueing scripts to copy files from /pds_san to the /scratch workarea.  Additionally, the PR includes jinja templating for the recipes, writes the thumbnail / browse image location to the existing json_keywords field, and adds search funtionality for thumbnail / browse queueing.

closes #441 
closes #442 
closes #453
closes #459 
closes #460 
closes #468
closes #471 
closes #472

Partially satisfies #465 requirement for mro_ctx and lro_lroc_edr recipe.


pds_pipelines/upc_process.py
```diff
@@ -285,7 +225,7 @@ def create_datafiles_record(label, edr_source, input_cube, session_maker):
     datafile_attributes['isisid'] = getISISid(input_cube)
 
     # Attempt to get the product id from the original label
-    product_id = getPDSid(input_cube)
+    product_id = getPDSid(input_cube + ".cub")
```
AustinSanders, line 228: This is the cause of the test failure.  I don't remember doing this, though.  I'll have to go back through and see what other changes might have been accidental.

acpaquette: Looks like you may just have to pull from upstream. I fixed this toward the beginning of the sprint and it looks like the old code is still in your PR

AustinSanders:
#472 turned out to include kind of a cool puzzle.  The queueing portion was super quick and easy, but getting derived_process to use a recipe that includes two gdal_translate steps was a little tricky.  JSON is parsed into nested dictionaries, and dictionaries require unique keys.  Since the recipe now includes two gdal_translate steps, the second gdal_translate was overwriting the first.

This was solved by creating a MultiDict object, which is a mostly normal dictionary, but it adds to list instead of overwriting on key collision.  I pointed the JSON parser at that object, so now, if there are duplicate steps, they are stored alongside their parameterizations.

pds_pipelines/utils.py
```diff
@@ -0,0 +1,146 @@
...
+    no_extension_inputfile = os.path.splitext(inputfile)[0]
+    cam_info_file = no_extension_inputfile + '_caminfo.pvl'
+    footprint_file = no_extension_inputfile + '_footprint.json'
+
+    template = jinja2.Template(recipe_string)
+    recipe_str = template.render(inputfile=inputfile,
+                                 no_extension_inputfile=no_extension_inputfile,
+                                 cam_info_file=cam_info_file,
+                                 footprint_file=footprint_file,
+                                 **kwargs)
```
acpaquette, line 49: This is something we probably should have touched base on. I was thinking we could solve this using a dictionary of properties like ALE is doing but the kwargs is basically just that. I think the kwargs is a better approach but the no_extension_inputfile, cam_info_file, and footprint_file should be part of those kwargs. Thus they should all be defined outside of this function

pds_pipelines/derived_queueing.py
```diff
@@ -69,15 +72,45 @@ def main():
...
+        size = 0
+        for element in qOBJ:
+            fname = path + element.filename
+            size += getsize(fname)
+
+        size_free = disk_usage(workarea).free
+        if size >= (disk_usage_ratio * size_free):
+            logger.error("Unable to process %s: size %d exceeds %d",
+                         volume, size, (size_free * disk_usage_ratio))
+            exit()
```
acpaquette, line 93: It would be a good idea to make this a function and add it to other queueing scripts in the future.

pds_pipelines/derived_process.py
```diff
@@ -0,0 +1,148 @@
+#!/usr/bin/env python
+import os
```
acpaquette, line 2: Removable imports:
- pvl
- sqlalchemy.orm.attributes import flag_modified

Removeable imports when functions get moved:
- datetime
- pytz
- Anything pysis

pds_pipelines/utils.py
```diff
@@ -0,0 +1,146 @@
+import os
+import jinja2
+import json
+import pds_pipelines
+import datetime
+import pytz
+from pds_pipelines.available_modules import *
```
acpaquette, line 7: Can this be removed?

AustinSanders: I don't think so.  This imports things like gdal_translate and pysis.isis, which are necessary for process()

pds_pipelines/derived_queueing.py
AustinSanders: This is kinda captured in #469 .  Are we ok with delaying this change until a new PR?

dpmayerUSGS: @AustinSanders Yes, I think we should address this, perhaps via #469, in a later PR. Thank you @acpaquette for the suggestion!

pds_pipelines/derived_process.py
```diff
@@ -0,0 +1,113 @@
...
+                recipe = json.load(fp, object_pairs_hook = parse_pairs)['reduced']
+                recipe_string = json.dumps(recipe['recipe'])
+
+            logger.info('Starting Process: %s', inputfile)
+
+            final_path = makedir(inputfile)
+            derived_product = os.path.join(final_path, os.path.splitext(os.path.basename(inputfile))[0])
+
+            no_extension_inputfile = os.path.splitext(inputfile)[0]
+            processes, workarea_pwd = generate_processes(inputfile,
```
acpaquette, line 87: Sorry to keep changing this but I don't think we even need to return workarea_pwd here. So far derived, map2, and pow just use the workarea variable from the config. I think we can just change upc to do the same and get ride of the workarea_pwd return.

pds_pipelines/derived_process.py
```diff
@@ -0,0 +1,113 @@
...
+            derived_product = os.path.join(final_path, os.path.splitext(os.path.basename(inputfile))[0])
+
+            no_extension_inputfile = os.path.splitext(inputfile)[0]
+            processes = generate_processes(inputfile,
+                                           recipe_string,
+                                           logger,
+                                           no_extension_inputfile=no_extension_inputfile, 
+                                           derived_product=derived_product)
+            failing_command = process(processes, workarea, logger)
+            # Ideally we could check for failing_command is None, but warnings count as errors
```
acpaquette, line 93: I'm not sure what you mean by warnings count as errors but don't we want to check if a failing command was returned? If one was we should error out and not update the database right?

dpmayerUSGS: ISIS (and GDAL) can print messages to `stderr` that are really just warnings, rather than errors that would preclude updating the database. Distinguishing between the two can be difficult.

AustinSanders: When I run derived_process.py, gdal_translate gives the warning "Warning 1: for band 1, nodata value has been clamped to 0, the original value being out of range."

The process completed and the output was as expected, but the warning was stored in failing_command.

PR #482, opened 2020-08-25
Loggy Fix

### LINKED ISSUE #477: Pow/map2 Unpopulated Loggy
##### Description

The loggy for both pow and map2 is no longer being populated as expected during pow and map2 processing. This breaks the **Log of commands file** being reported when pow and map2 jobs are finished. See [this job](https://astrocloud.wr.usgs.gov/index.php?view=viewjob&key=c8c5e0f107f1e5f0469a14ce4ebe1b4b) as an example of what the file should look like.

This was previously being done during processing, where each job after processing would be submitted to the loggy with the arguments run and the status of the command (success or failure). There was also a link provided by each app to get a more detailed explanation of what said app does.

With the new process the loggy is no longer being populated and is currently being submitted to the loggy queue as an empty dictionary. Ideally, what is being placed on the loggy queue should remain the same as before.

##### Proposed Solution

Reproduce what the loggy was producing before only using python dictionaries, and populate the loggy queue with the same information.

acpaquette:
Fixes #477 where the logging broke after PR #473 

A quick overview is that the processes are collected into dictionary detailing every successfully run command in the pipeline, along with a help link for each command. This fix also populates the loggy queue appropriately when a POW or MAP2 job finishes.

See issue #477 for more details on the bug.

### DIFF
--- a/pds_pipelines/upc_process.py
+++ b/pds_pipelines/upc_process.py
@@ -491,7 +491,7 @@ def main(user_args):
                                        no_extension_inputfile=no_extension_inputfile,
                                        cam_info_file=cam_info_file,
                                        footprint_file=footprint_file)
-        failing_command, error = process(processes, workarea, logger)
+        failing_command, _ = process(processes, workarea, logger)
 
         pds_label = pvl.load(inputfile)
 
--- a/pds_pipelines/upc_process.py
+++ b/pds_pipelines/upc_process.py
@@ -491,7 +491,7 @@ def main(user_args):
                                        no_extension_inputfile=no_extension_inputfile,
                                        cam_info_file=cam_info_file,
                                        footprint_file=footprint_file)
-        failing_command = process(processes, workarea, logger)
+        failing_command, error = process(processes, workarea, logger)
 
         pds_label = pvl.load(inputfile)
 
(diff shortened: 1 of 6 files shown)


pds_pipelines/upc_process.py
```diff
@@ -491,7 +491,7 @@ def main(user_args):
                                        no_extension_inputfile=no_extension_inputfile,
                                        cam_info_file=cam_info_file,
                                        footprint_file=footprint_file)
-        failing_command = process(processes, workarea, logger)
+        failing_command, error = process(processes, workarea, logger)
```
dpmayerUSGS, line 494: @acpaquette What does this change do?

acpaquette: The process function now returns a tuple of the failing command along with its error message if something failed. To keep what was expected to return from process we need to unpack the return from `process`. I am going to change `error` to `_` so that the return is a little more clear.

dpmayerUSGS: Cool! 
Will this change mean that the failing command and the error message will be inserted into the `jsonkeywords` field of the `json_keywords` table? Currently, only the failing command is inserted.

acpaquette: No, only the failing_command will be placed into the json_keywords. I can update it to add both but that may be tangential to this PR

dpmayerUSGS: No need to update; I was just curious. I concur that such an update would be tangential to this PR.

PR #494, opened 2020-09-02
Include .aux.xml and .msk in final POW/MAP output. Minor config updates

### LINKED ISSUE #489: PNG, Jpeg runs for MAP2 and I bet POW are not copying over worldfile or auxiliary sidecar files
Based on the results for this job,
https://astrocloud.wr.usgs.gov/index.php?view=viewjob&key=b3a0e5f5a18d9ff0d26e4f3dbf8a80bd

I am no longer seeing the geospatial side-car files as saved by gdal_translate for formats like Jpeg, PNG, ... This would be either a worldfile (*.jgw, *.pgw, *.gfw, *.wld), a GDAL auxiliary file (*.aux), or both. Since this are critical for GIS applications and they are small, they should be copied with the output images and *.map file (and available in the download zip).

### LINKED ISSUE #491: derived_process.py: Not capturing output of process() correctly causes failure
The `process()` function was recently changed to return a 2-item tuple rather than a single item, but `derived_process.py` wasn't updated to capture this correctly. This causes `derived_process.py` to think that a command in the derived recipe has failed, and exit without writing the URL of the derived image to the UPC database.

The following line should be updated to received `failing_command, _`
https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/34225dec1b1f2cb484892b1604bb03a9d16aadf2/pds_pipelines/derived_process.py#L92

### LINKED ISSUE #492: PDSinfo.json: Add mars_science_laboratory
Add an entry to `PDSinfo.json` for the Mars Science Laboratory archive. This archive is *not* UPC-eligible and does *not* need an associated recipe file. It is treated as a monolithic archive (rather than being split into different product types) because it only needs to be ingested and DI'd.

```
    "mars_science_laboratory": {
        "archiveid": "129",
        "path": "/pds_san/PDS_Archive/Mars_Science_Laboratory/"
    },
```

### LINKED ISSUE #493: jobconfig.py: Replace browse/thumbnail with "derived" SBATCH template
With the consolidation of browse/thumbnail processing into a single "derived" process, the SBATCH templates in `jobconfig.py` need to be updated so that `derived_process.py` can be launched in SLURM via `hpc_job_manager.py` This is a bug because the job manager can't launch derived process jobs without this change.

The new derived SBATCH template can use the same parameters as the existing browse template, but with names of log/script files (etc.) updated to use "derived" rather than "browse."

The browse and thumbnail templates in `jobconfig.py` should be deleted.

dpmayerUSGS:
This PR primarily addresses #489. Any .aux.xml or .msk files that are created during POW or MAP2 processing are now included in the final output. The code changes streamline packaging of final POW/MAP2 output by adding ancillary files (.aux.xml and .msk), not just the final image, to the zip queue in the process scripts. `service_final.py` can then simply loop over the items in the zip queue and add them to the zip file. The subprocess calls to `zip` have been replaced with the Python library `zipfile`. This removes a tacit dependency on a command line tool that is not guaranteed to be available on a system where PDS-Pipelines is installed.
Fixes #489 

Inserted a missing `, _` in `derived_process.py`
Fixes #491  

Added an entry for MSL to the PDS info dict to support ingest/DI of this archive
Fixes #492 

Updated SBATCH templates to support derived_process, deleted deprecated browse and thumbnail templates. Tweaked some SLURM job params.
Fixes #493


USGSsakins (approved):
The code looks much cleaner ... nice job.

PR #501, opened 2020-09-17
POW recipe updates, recipe cleanup, and service_job_manager.py bug fixes

### LINKED ISSUE #260: Clean Up Recipe Files
The `recipe` directory and its subdirectory `new` together hold > 150 recipe files, many of which appear to be effectively empty (contain skeletons of valid recipes, but none of the values are populated), refer to data sets that don't exist, or are duplicates of other recipes (but listed under an incorrect name).

We should start cleaning these up by cross-referencing recipe files with entries in `PDSinfo.json`. Recipe files that do not have a corresponding entry in the PDS info dictionary are good candidates for deletion.

Recipe files that are effectively empty or are duplicates of valid, operational recipes are also good candidates for deletion.

If a recipe file that is effectively empty, but has a corresponding entry in `PDSinfo.json`, may need to be filled in, in which case a new issue should be opened to add it.

There are also recipes that may need to be modified to make them valid (that is, they run correctly, as intended) or to remove unnecessary elements (for example, not all of the items in a `upc_reqs` list may be necessary for a given dataset).

### LINKED ISSUE #496: ingest_process.py: Change log level about empty upc_reqs from .warning to .debug
https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/46004032aeb91498c55a34b91a0211577072dd7f/pds_pipelines/ingest_process.py#L104

In typical usage, this message just ends up cluttering the ingest log file. It is perfectly valid for datasets that are not UPC-eligible to not have `upc_reqs` defined. Raising the log level of this message to debug will help keep the log file more concise in typical usage, while still allowing easy access to this information during testing and debugging.

### LINKED ISSUE #498: Update POW Recipes of Older Datasets to Use Jinja Templates
#465 Covered updating POW recipes for relatively new datasets. This issue is basically the same, but applies to the remaining datasets that are currently supported in PILOT as POW-eligible, and (with the exception of HRSC) come from missions that concluded >25 years ago.

**Definition of done**
All the POW recipes for datasets in the checklist below are modified to use the jinja tokens as appropriate, following the example of the existing POW recipes. 
 - [x] clementine_hires_edr
 - [x] clementine_nir_edr
 - [x] clementine_uvvis_edr
 - [x] mariner_10
 - [x] mars_express
 - [x] voyager

### LINKED ISSUE #499: service_job_manager.py: getCleanName() skips candidates while trying to test for upc_reqs
**Description**
The logic for filtering out candidate clean names in `getCleanName()` is broken in a way that causes it to erroneously skip trying to evaluate one or more candidates after removing a candidate.

Specifically, if there is more than 1 candidate, the function enters a for loop to test whether the `upc_reqs` items are found in `file_name`:
https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/46004032aeb91498c55a34b91a0211577072dd7f/pds_pipelines/service_job_manager.py#L67-L77

However, `candidates.remove(item)` decrements the index of the remaining items in the list`candidates` so that the next pass through the for loop skips what "should" be the next item to be tested. Thus, if `candidates = ["foo", "bar", "baz"]`, and `"foo"` is removed, on the next pass through the for loop, `item = "baz"` because now `candidates[0] =  "bar"` and `candidates[1] =  "baz"`. If `"baz"` is not removed, then there are 2 items left in `candidates` and the function raises an error that it can't unambiguously identify the clean name.

**Proposed Solution**
Rather than removing items from `candidates`, the test condition on Line 69 should be negated, and `item` should be *appended to a new list* (perhaps named `filtered_candidates`), rather than be removed from `candidates` . The function would still go on to test if `len(filtered_candidates) > 1`, and raise an error if true. It's still considered valid for `upc_reqs` to be absent from an entry in `PDSinfo.json`, so rather than `pass`ing on a KeyError like the code does now, the `item` should be appended to the filtered candidate list.

**Additional Information**
As noted in the comments inside the function, `getCleanName()` is a nasty hack. Replacing this function with some other way of obtaining the archive clean name (or finding an alternative to the archive clean name concept altogether) is outside the scope of this issue.

### LINKED ISSUE #500: service_job_manager.py: Be more tolerant when pvl-style label cannot be read from inputfile
**Description**
https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/46004032aeb91498c55a34b91a0211577072dd7f/pds_pipelines/service_job_manager.py#L990-L991

The `generate_pow_recipe()` function currently requires a PDS label as one of its inputs in order to test if certain dataset-specific conditions exist that require the recipe skeleton to be modified.

For so-called "pre-PDS" datasets such as Voyager and Viking, which are officially PDS archives, but stored in legacy formats (usually with some form of compression), it is not possible to extract a PDS3 label from them using `pvl`. These datasets don't (shouldn't?) require further introspection in order to modify recipes on-the-fly, so requiring a PDS label as one of the inputs to `generate_pow_recipe()` is unnecessarily strict.

This is a bug because if `pvl.load()` fails to load a label on Line 990 above, it causes the entire script to fail with an error coming from `pvl`. The exact type of error doesn't really matter, but it usually has to do with `pvl` misinterpreting characters near the beginning of the affected file. 

**Proposed Solution**
- Wrap the `pvl.load()` call in a try/except block and maybe print an informational message to the log if an error is raised, but allow the script to continue.
- make `pds_label` an optional input to `generate_pow_recipe()`, only entering the dataset-specific test block if `pds_label` is not `None`

dpmayerUSGS:
Deleted unused/empty recipe files from `recipe` directory. Moved active recipe files up one directory from `recipe/new/`. After deleting extraneous recipe files, there is no need to maintain the `new` subdirectory of `recipe`, so deleted the subdirectory. Changed default value of `recipe_base` in `config.py` accordingly.
Fixes #260

Updated POW recipes for several datasets to use jinja templates. Made corresponding tweaks to `PDSinfo.json`.
Tested the recipes by finding images for each dataset in PILOT, submitting jobs, then calling `service_job_manager.py` from a test environment. Confirmed that recipes produce expected output.
Fixes #498

Corrected logic in `service_job_manager.py` for determining which archive clean name to use for a particular job
Fixes #499

Made `service_job_manager.py` tolerant to cases where the input file is in a pre-PDS format (i.e. Viking orbiter images) and `pvl` cannot extract a label. Rather than causing the script to fail with an error, it will now attempt to continue processing the input file, which (barring some additional, unrelated problem with the input file) should allow it to be processed normally.
Fixes #500

Changed log level of a message in `ingest_process.py`
Fixes #496

During testing, discovered that a call to `.write()` in `make_map.py` needed encoded bytes as input rather than stream.


USGSsakins (approved):
This looks good.  It's helpful to clear out these extraneous files from the repo.

PR #503, opened 2020-10-14
make_map.py: Change how PVL of map templates is encoded

### LINKED ISSUE #502: make_map.py: Upstream change in PVL breaks map template formatting
**Description**
This is a rather esoteric issue that may not strictly be a bug in PDS-Pipelines so much as it is a weird interaction between a [recent change in the default behavior of `pvl.dump()`](https://pvl.readthedocs.io/en/latest/encoding.html#pre-1-0-pvl-dump-behavior) and the ISIS `cam2map` program.

The `MakeMap` class is used to build a map template file for POW/MAP2 jobs by constructing a JSON dict, converting this to a PVL object and writing to a file. In `pvl` 1.x, the default behavior of `pvl.dump()` and `pvl.dumps()` is to write PVL text that is strictly PDS3 compliant, which involves forcing all keywords to UPPER CASE, so `MakeMap()` creates templates that look like
```
GROUP            = Mapping
PROJECTIONNAME   = PolarStereographic
PIXELRESOLUTION  = 50.0
TARGETNAME       = MOON
EQUATORIALRADIUS = 1737400.0
POLARRADIUS      = 1737400.0
END_GROUP        = Mapping
END
``` 

For POW jobs, this template is then passed as one of the arguments to `cam2map`, which runs without error. However, the projection name is *missing* from the `Mapping` group in the resulting map-projected cube:
```
 Group = Mapping
    CenterLongitude    = 170.80390729546
    LatitudeType       = Planetocentric
    LongitudeDirection = PositiveEast
    LongitudeDomain    = 360
    MinimumLatitude    = 76.802853123529
    MaximumLatitude    = 78.602063501845
    MinimumLongitude   = 170.57938691151
    MaximumLongitude   = 171.02842767942
    UpperLeftCornerX   = -1600.0 <meters>
    UpperLeftCornerY   = -346750.0 <meters>
    PixelResolution    = 50.0 <meters/pixel>
    Scale              = 606.46700848299 <pixels/degree>
    CenterLatitude     = 90.0
  End_Group
```

 The image itself looks as expected, but GDAL doesn't recognize the spatial reference system, so if the job recipe calls `gdal_translate` to convert to another format, the result will have *no* geospatial information. 

**Proposed Solution**
Change https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/eb0a81d42e4ef0ca3f0a7d61c6a68c61c316bf4c/pds_pipelines/make_map.py#L182-L185
to explicitly invoke the ISIS encoder in `pvl`:
```
        pvl.dump(self.mapDICT, filename, encoder=pvl.encoder.ISISEncoder())
```

I may open a related issue on USGS-Astrogeology/ISIS as well because it seems like `cam2map` shouldn't be case sensitive about map templates *or* it should be throwing an error if the map template doesn't meet the required format.

dpmayerUSGS:
Specifically invoke the ISIS encoder to ensure that keywords in the map template are camel case instead of all caps.
Tested on POW job [c961f25e1e943432809457db16ad9e95](https://astrocloud.wr.usgs.gov/index.php?view=viewjob&key=c961f25e1e943432809457db16ad9e95) and verified that both the resulting map template and final projected GeoTIFF are correct.

Fixes #502

### DIFF
--- a/pds_pipelines/make_map.py
+++ b/pds_pipelines/make_map.py
@@ -157,7 +157,7 @@ class MakeMap(object):
         """
         self.mapDICT['End_Group'] = 'Mapping'
 
-        mappvl = pvl.dumps(self.mapDICT)
+        mappvl = pvl.dumps(self.mapDICT, encoder=pvl.encoder.ISISEncoder())
         self.mappvl = mappvl
 
         return mappvl
@@ -179,7 +179,4 @@ class MakeMap(object):
         filename
         """
         self.mapDICT['End_Group'] = 'Mapping'
-        tempPVL = pvl.dumps(self.mapDICT)
-
-        with open(filename, 'wb') as f:
-            f.write(tempPVL.encode())
+        pvl.dump(self.mapDICT, filename, encoder=pvl.encoder.ISISEncoder())


thareUSGS:
Looks good to me. Nice that encoder=pvl.encoder.ISISEncoder does the camel case for you (I assume).

USGSsakins (approved):
Looks good.

PR #508, opened 2020-11-06
Switch link_artifacts.py to get paths from PDSinfo.json, plus miscellaneous small bugfixes

### LINKED ISSUE #495: DeprecationWarning: The 'warn' method is deprecated, use 'warning' instead
Apparently `.warn` is deprecated in favor of `.warning`. This is causing `DeprecationWarning` to be raised in SLURM error logs when a warning message is printed to logger.

`logger.warn()` should be replaced with `logger.warning()` throughout the following files:

- [ ] `di_process.py`
- [ ] `di_queueing.py`
- [ ] `ingest_process.py`
- [ ] `ingest_queueing.py`
- [ ] `link_artifacts.py`
- [ ] `service_job_manager.py`

### LINKED ISSUE #504: pow_process.py: Not splitting file name correctly when output is cube
If a user specifies ISIS cube as the output format for a POW job *and* requests a simple stretch or changes bit type, `pow_process.py` is including the "plus attribute" string in the final file name, which results in the final file not being found (because it doesn't exist).
https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/eb366d09d7fe57d67ed8372b5ccae52b93a502da/pds_pipelines/pow_process.py#L102

`pow_process.py` *should* split the name on the '+' character as is done in `map_process.py`

### LINKED ISSUE #505: link_artifacts.py: use PDSinfo.json to get archive path rather than recipe file
This script to create symbolic links to datasets is normally run following `ingest_queueing.py`. It currently tries to look up a recipe JSON file associated with the dataset mnemonic listed in the Redis LinkQueue. However, not every dataset listed in PDSinfo.json has a corresponding recipe file because recipe files are really only applicable for UPC-eligible datasets.

This is a bug because when trying to run `link_artifacts.py` when the LinkQueue contains non-UPC-eligible datasets, the script fails.

`link_artifacts.py` should be modified to get the `path` keyword for the given dataset from PDSinfo.json, rather than the `src` keyword from a (possibly non-existent) recipe file.

### LINKED ISSUE #506: link_artifacts.py: Replace boilerplate description with actual description
https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/eb366d09d7fe57d67ed8372b5ccae52b93a502da/pds_pipelines/link_artifacts.py#L15

### LINKED ISSUE #507: hpc_job.py: Remove `--spread-job` SBATCH parameter
https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/eb366d09d7fe57d67ed8372b5ccae52b93a502da/pds_pipelines/hpc_job.py#L164

We added this in a while ago when we thought that forcing array jobs to be spread over many nodes was necessary/desirable. (In fact, it didn't do anything at all because until last week, we weren't running a version of SLURM that was new enough to support the option anyway.) Forcing jobs to be spread out like this isn't necessary; SLURM is capable of distributing jobs in a way that is reasonable for PDS-Pipelines processes. 

This line can simply be deleted.

dpmayerUSGS:
Changed `link_artifacts.py` to read path to archive from `PDSinfo.json` instead of individual recipe files (Fixes #505), plus updated description in argparse help (Fixes #506).

pow_process.py now correctly splits filenames on '+' character in cases when output format is an ISIS cube and there are other attributes specified, such as a simple stretch or a change in bit type (Fixes #504).

Replaced instances of `.warn()` with `.warning()` throughout code (Fixes #495). Removed `--spread-job` SBATCH parameter from `hpc_job.py` so HPC jobs will no longer use it (Fixes #507).

### DIFF
--- a/pds_pipelines/di_process.py
+++ b/pds_pipelines/di_process.py
@@ -69,7 +69,7 @@ def main(user_args):
             Qelement = session.query(Files).filter(
                 Files.filename == inputfile).one()
         except Exception as e:
-            logger.warn('Filename query failed for inputfile %s: %s', inputfile, str(e))
+            logger.warning('Filename query failed for inputfile %s: %s', inputfile, str(e))
             continue
 
         archive_path = PDSinfoDICT[archive]['path']
@@ -84,7 +84,7 @@ def main(user_args):
 
             Qelement.di_pass = checksum == Qelement.checksum
             if !Qelement.di_pass:
-                logger.warn('File %s checksum %s does not match the database entry checksum %s', 
+                logger.warning('File %s checksum %s does not match the database entry checksum %s', 
                             cpfile, checksum, Qelement.checksum)
 
             Qelement.di_date = datetime.datetime.now(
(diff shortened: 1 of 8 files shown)


USGSsakins (approved):
Looks good.

PR #517, opened 2021-01-07
Factored out isis_sn from derived_process

### LINKED ISSUE #512: derived_process.py: ISIS serial number unreliable for relating file to upcid
## Description
When running `derived_process.py`, certain input files are matched to the wrong entry in the UPC database.

## Expected Behavior
Running `derived_process.py` should always match the input file to the correct entry in the UPC database

## Suspected Cause
`derived_process.py` computes the ISIS serial number of a given input file and then queries the DataFiles table in the UPC database for that serial number, and assumes that the upcid of the first result corresponds with the record for that file. However, ISIS serial numbers are not guaranteed to be unique within a dataset, so an input file may be matched to the wrong upcid.
https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/391d7fab51c00a50b3b948e0313f13dc9767add0/pds_pipelines/derived_process.py#L96

## Additional Context
For *most* datasets, ISIS serial numbers ("isisid" in the UPC database schema) are unique within the dataset and therefore can be used to relate a particular file back to a upcid in the UPC database. However, they are *not* unique for files within archives including (but not limited to?) Clementine and Galileo SSI. In the case of Clementine (HiRES, NIR, or UVVIS), distinct images can have the same ISIS serial number. 

For example, [LHA1365J.328](https://pdsimage.wr.usgs.gov/Missions/Clementine/cl_0077/lun328/lhxxxxxx/lhxxxxxj/lha1365j.328) and [LHA1366J.328](https://pdsimage.wr.usgs.gov/Missions/Clementine/cl_0077/lun328/lhxxxxxx/lhxxxxxj/lha1366j.328) both yield the ISIS serial number "Clementine1/HIRES/1994-04-29T08:48:21.170" 

## Proposed Solution
The only thing that really makes a UPC record unique is the *path to the source file*. `derived_process.py` should compute the expected path to the source file based on file it is generated browse/thumbnail images for, and query the "source" column in the DataFiles table instead of the "isisid" column to try and fetch the upcid that corresponds to the input file.

AustinSanders:
Closes #512

### DIFF
--- a/pds_pipelines/derived_process.py
+++ b/pds_pipelines/derived_process.py
@@ -12,8 +12,8 @@ from pds_pipelines.redis_queue import RedisQueue
 from pds_pipelines.redis_lock import RedisLock
 from pds_pipelines.db import db_connect
 from pds_pipelines.models.upc_models import DataFiles, SearchTerms, JsonKeywords
-from pds_pipelines.config import pds_log, pds_info, workarea, pds_db, upc_db, lock_obj, upc_error_queue, recipe_base, archive_base, derived_base, derived_url
-from pds_pipelines.utils import generate_processes, process, parse_pairs, add_process_db, get_isis_id
+from pds_pipelines.config import pds_log, pds_info, workarea, pds_db, upc_db, lock_obj, upc_error_queue, recipe_base, archive_base, derived_base, derived_url, web_base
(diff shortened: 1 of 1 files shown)


dpmayerUSGS (approved):
Looks good. 
+1 for changing fuzzy search to exact match.

PR #518, opened 2021-01-07
Added default nodata value

### LINKED ISSUE #474: derived_process recipes: gdal_translate produces nodata warning
When running derived_process.py, gdal_translate gives the warning "Warning 1: for band 1, nodata value has been clamped to 0, the original value being out of range."

This occurs when the input to gdal_translate, normally a 32bit file is flattened to 8bit but the recipe does not explicitly reassign the NoData value. This is only a warning, so gdal_translate still produces the desired output, but it's kind sloppy to run commands that always produce unhelpful warning messages. I think the solution is to add the Python equivalent of `-a_nodata 0` to the gdal_translate command in each recipe. 

_Based on comment from @AustinSanders in https://github.com/USGS-Astrogeology/PDS-Pipelines/pull/470_

AustinSanders:
closes #474

### DIFF
--- a/pds_pipelines/available_modules.py
+++ b/pds_pipelines/available_modules.py
@@ -34,6 +34,12 @@ def gdal_translate(dest, src, *args, **kwargs):
         # If outputType not specified, no conversion is necessary and GDAL will
         #  use default arguments.
         pass
+    try:
+        # check for existing noData specification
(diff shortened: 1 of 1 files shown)


dpmayerUSGS (approved):
Looks good. Nice straightforward solution.

PR #519, opened 2021-01-07
Update appropriate queues/hashes with error messages

### LINKED ISSUE #513: POW/MAP2: Error message not being set in cluster jobs database
## Description 
If a processing step in a POW or MAP2 job generates an error (i.e. `cam2map` failed), `service_final.py` is supposed to insert text of the error message(s) to the "error" field in the "processing" table of the cluster jobs database. At some point in the last 6 months, this functionality got removed (probably when other improvements to error logging were being made). This is a bug because `notify_error()` in `notify.py` relies on this flag being set in order to send out an email to the user that their job is finished *and* that it included one or more errors.

It *looks like* the following line is supposed to be inserting the error message to the database, but it is quite obviously not working correctly:
https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/391d7fab51c00a50b3b948e0313f13dc9767add0/pds_pipelines/service_final.py#L114

The effect of this now is that the user only ever receives the "normal" email notification that their job is finished, and it is up to the user to look at the log file (or notice that the number of output files doesn't equal the number of input files) to realize that anything is amiss.

## Expected Behavior
If a user submits a POW or MAP2 job that generates processing errors, the error text should get set in the processing table so that the user is then sent the correct type of email notification.

acpaquette:
Adds back in error queueing that was removed in a refactor some time ago. This means that errors should appropriately be queued for the associated map2/pow jobs if/when they error.

Closes #513

### DIFF
--- a/pds_pipelines/map_process.py
+++ b/pds_pipelines/map_process.py
@@ -8,7 +8,7 @@ import logging
 import shutil
 import argparse
 
-from pds_pipelines.config import lock_obj, pds_log, default_namespace, workarea
+from pds_pipelines.config import lock_obj, pds_log, upc_error_queue, default_namespace, workarea
 from pds_pipelines.redis_queue import RedisQueue
 from pds_pipelines.redis_lock import RedisLock
 from pds_pipelines.redis_hash import RedisHash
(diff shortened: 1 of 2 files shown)


AustinSanders (approved):

dpmayerUSGS (approved):
Looks good

PR #520, opened 2021-01-08
Allow users to import the library with no active DB connection

### LINKED ISSUE #386: Unable to import library without a DB running
It looks like someone put in logic that should allow for the import of a module in pds_pipelines without a DB. Looking in db.py it looks like the original intent was to attempt to create a connection and if the creation failed to return None, None. I think this is the case because lines [17-18](https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/master/pds_pipelines/db.py#L17) are setting the variables to default to None.

It looks like the later modified such at the config has defaults inside the library. This makes is so that the try will never fail. If the try never fails, then the check to see if a valid session is returned in [models/upc_models.py](https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/master/pds_pipelines/models/upc_models.py#L156) always returns a valid session object, even if no DB exists.

So the scaffolding is there to not need a DB to even import, but it looks like this has been slowly (and unintentionally) modified away.

AustinSanders:
returns none/none (sessionmaker / engine) if the the engine does not yield a valid connection.

I'm now able to use:
import pds_pipelines
from pds_pipelines import *
from pds_pipelines.models import upc_models
from pds_pipelines.models import pds_models

Closes #386

### DIFF
--- a/pds_pipelines/db.py
+++ b/pds_pipelines/db.py
@@ -32,6 +32,12 @@ def db_connect(cred):
                                                                     c[cred]['db']))
     except KeyError:
         print("Credentials not found for {}".format(cred))
+
+    try:
+        engine.connect()
+    except sqlalchemy.exc.OperationalError:
+        print(f"WARNING:  Unable to create a database connection for credential specification '{cred}'")
+        return None, None
     metadata = MetaData(bind=engine)
     session_maker = sessionmaker(bind=engine)
     return session_maker, engine


dpmayerUSGS (approved):
Looks good to me.
+1 for adding the warning

acpaquette (approved):
I tested this locally and things seem good. I'm not sure what is going on with travis

PR #523, opened 2021-01-14
Make sql operations retry on failure.

### LINKED ISSUE #514: Make UPC/Derived Processing More Robust to Database Connection Errors
I'm going to frame this as more of a user story than a specific issue with an obvious closed form solution.

## Background
1. Several scripts in PDS-Pipelines use Redis queues to conduct items through processing tasks.
2. Several processing scripts in PDS-Pipelines pop items off of Redis queues, operate on files based on the content of the queued item, and write information to various databases when they're done with file processing.
3. The machines that host the various databases that the production installation of PDS-Pipelines connect to sporadically and unpredictably go down (usually briefly, often for just a few seconds), which causes the processing script that was trying to connect to a given database to fail with an error.

## The Thing That's Happening That I Don't Like
It's never great when a script fails with an error, but there is an added complication for PDS-Pipelines due to the way queued items are handled: If, say in `upc_process.py` a database insert fails because the script can't connect to the UPC database, not only does the script fail, but information about what item from the Redis queue that the script was working on is lost. This is a problem because as a user, all I know is that one of the instances of `upc_process.py` failed due to a database connection error, but I don't know which item from the Redis queue it had been working on at the time, so I don't know what file I need to re-queue and try again. I *might* be able to reconstruct which file was being worked on when the script failed by carefully cross-referencing the SLURM log and `UPC.log`, but this is tedious and many of the users who are actually running UPC processing don't have the requisite skills to do this.

## As a user...
1. I would like `upc_process.py`, and `derived_process.py` to catch and log exceptions associated with database connection issues, rather than fail hard
2. I would like `upc_process.py`, and `derived_process.py` to automatically re-add an item to the associated Redis queue if the script encounters a database connection issue

## Possible Solution
 `upc_process.py` and `derived_process.py` already parse items coming off a particular Redis queue (obviously). Perhaps a solution is as simple as storing the contents of that item verbatim in memory as soon as it's popped off the queue, then making sure that any functions performing database queries/inserts are wrapped in try/except blocks, and then pushing the item back to the Redis queue if an exception is raised.

AustinSanders:
Closes #514

### DIFF
--- a/pds_pipelines/db.py
+++ b/pds_pipelines/db.py
@@ -2,11 +2,11 @@ from sqlalchemy import create_engine, inspect
 from pds_pipelines.config import credentials as c
 from sqlalchemy import MetaData, Table
 from sqlalchemy.orm import mapper, sessionmaker
-from sqlalchemy.exc import OperationalError
+import sqlalchemy
 
 def db_connect(cred):
     """
-    Given some crediential string that is present as a key in the config
+    Given some credential string that is present as a key in the config
     credentials dictionary, generate the engine and associated session_maker
     objects for a database.
 
--- a/pds_pipelines/db.py
+++ b/pds_pipelines/db.py
@@ -33,7 +33,7 @@ def db_connect(cred):
                                                   c[cred]['port'],
                                                   'postgres')
 
-        engine = create_engine(public_database_url)
+        engine = create_engine(public_database_url, pool_pre_ping=True)
     except KeyError:
         print("Credentials not found for {}".format(cred))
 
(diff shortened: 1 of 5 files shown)


pds_pipelines/db.py
```diff
@@ -33,7 +33,7 @@ def db_connect(cred):
                                                   c[cred]['port'],
                                                   'postgres')
 
-        engine = create_engine(public_database_url)
+        engine = create_engine(public_database_url, pool_pre_ping=True)
```
acpaquette, line 36: Just curious, what does his change?

AustinSanders: The documentation isn't great, but when you enable pool_pre_ping, anything using a session "pre pings" the session to ensure that it's still alive.  I _think_ that it attempts to reconnect / create a new session if the session has died.

acpaquette (approved):
Quick question, overall looks solid!

PR #524, opened 2021-01-16
UPCkeyword Removal

### LINKED ISSUE #412: Clean up/Remove UPCkeyword Class
The UPCKeyword class in upc_keyword.py is a thin wrapper around a pvl load call and a recursive search through a pvl object. This doesn't need to exist as a class and could be broken down to stand along functions.

acpaquette:
Moves the UPCkeywords object to stand alone functions in a util module. Helps to simplify things and remove an unnecessary object layer.

Closes #412


acpaquette:
@dpmayerUSGS It's possible that with updating to PVL 1.0 we can remove the load function in this entirely. The new grammer module handles more delimeteres [see here](https://github.com/planetarypy/pvl/blob/main/pvl/grammar.py) so we don't have to do these replacements on the contents of the file anymore.

AustinSanders (approved):

PR #526, opened 2021-01-19
Gdal Unscale Update

### LINKED ISSUE #511: bit type conversion (scale and offset applied) going to 32bit?
John B. was helping a user with a MAP2 job and perhaps supporting a new default behavior might be warranted? 

First, this is a 16bit file with an offset and scale and the user requested 32bit float (real). While there was a conversion to 32bit during gdal_translate, the offset and scaler where never applied. In the case for 16bit -> 32bit this probably does make sense to do by default. Note ISIS 32bit cubes do not allow for scalar and offset, and thus any unscale must be applied.

I'm not sure if this can be handled in map2map or will need its own ISIS bit conversion routine. Otherwise, I guess "-unscale" could be used during the gdal_translate conversion.

acpaquette:
Updated map and pow jobs to unscale if the output bit type is a real. This also fixes the gdal and ogr imports for the available modules.

Closes #511

### DIFF
--- a/pds_pipelines/available_modules.py
+++ b/pds_pipelines/available_modules.py
@@ -10,7 +10,7 @@ except Exception as e:
                  f'Failed with the following {e}.')
 
 try:
-    import gdal
+    from osgeo import gdal
     available_modules['gdal'] = gdal
     # Python GDAL doesn't use exceptions by default.
     gdal.UseExceptions()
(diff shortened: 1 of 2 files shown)


dpmayerUSGS (approved):
+1 for updating the gdal/ogr import statements

PR #527, opened 2021-01-20
Standardized/modularized queues

AustinSanders:
First pass at standardizing / modularizing queues.  Needs testing and minor reworks.

The first pass includes an attempt at establishing an interface for queueing processes.  Since it was reverse-engineered, it's somewhat of a loose-fitting interface at the moment.  I'd like to see if/how we can polish it.

The QueueProcess class serves as a base process model, and it stores all the information necessary to get the queueing processes up and running (archive, volume, search string information, a session maker, etc.).  Each of the child classes is responsible for providing a "get_matching_files" and "enqueue" implementation, and that seems to be enough to get things moving along.

I wasn't sure about storing the session maker and Redis connections within the objects, but I decided to allow the object to self-manage / maintain its connections because they are required for proper queueing operation, i.e. it doesn't ever make sense to have a QueueProcess that doesn't have a database connection or Redis connection (since we're grabbing items from the database and pushing them to a Redis queue).

Things I need to do before this is ready for review / merge:
- [x] A full run of DI processing
- [x] A full run of UPC processing
- [x] A full run of derived processing
- [x] Additional docstrings
- [x] Removal of unused imports

### DIFF
--- a/pds_pipelines/available_modules.py
+++ b/pds_pipelines/available_modules.py
@@ -10,7 +10,7 @@ except Exception as e:
                  f'Failed with the following {e}.')
 
 try:
-    import gdal
+    from osgeo import gdal
     available_modules['gdal'] = gdal
     # Python GDAL doesn't use exceptions by default.
     gdal.UseExceptions()
--- a/pds_pipelines/queueing.py
+++ b/pds_pipelines/queueing.py
@@ -3,6 +3,8 @@ import json
 import logging
 import pathlib
 import glob
+import datetime
+import pytz
 
 from shutil import copy2, disk_usage
 from sqlalchemy import Date, cast
--- a/pds_pipelines/queueing.py
+++ b/pds_pipelines/queueing.py
@@ -65,7 +65,6 @@ def copy_files(fname, src_dir, dest_dir):
     dest_path = dest_path.replace(src_dir, dest_dir)
     pathlib.Path(dest_path).mkdir(parents=True, exist_ok=True)
     for f in glob.glob(splitext(fname)[0] + r'.*'):
-        print(f)
         copy2(f, dest_path)
     return join(dest_path, basename(fname))
 
@@ -149,15 +148,17 @@ class QueueProcess():
         logger : logging.Logger
             The parameterized logger
         """
-        logger = logging.getLogger(f"{self.process_name}_Queueing.{self.archive}")
-        level = logging.getLevelName(log_level)
-        logger.setLevel(level)
-        logFileHandle = logging.FileHandler(pds_log + 'Process.log')
-        formatter = logging.Formatter(
-            '%(asctime)s - %(name)s - %(levelname)s, %(message)s')
-        logFileHandle.setFormatter(formatter)
-        logger.addHandler(logFileHandle)
-        return logger
+        if not hasattr(self, "logger"):
+            logger = logging.getLogger(f"{self.process_name}_Queueing.{self.archive}")
+            level = logging.getLevelName(log_level)
+            logger.setLevel(level)
+            logFileHandle = logging.FileHandler(pds_log + 'Process.log')
+            formatter = logging.Formatter(
+                '%(asctime)s - %(name)s - %(levelname)s, %(message)s')
+            logFileHandle.setFormatter(formatter)
+            logger.addHandler(logFileHandle)
+            self.logger = logger
+        return self.logger
 
 
     def get_matching_files(self):
--- a/pds_pipelines/queueing.py
+++ b/pds_pipelines/queueing.py
@@ -40,7 +40,6 @@ def parse_args():
     return args
 
 
-
 def copy_files(fname, src_dir, dest_dir):
     """ Globs files with shared filename in src_dir and copies them to dest_dir
 
@@ -70,6 +69,31 @@ def copy_files(fname, src_dir, dest_dir):
 
 
 def has_space(elements, src_path, dest_path, ratio):
+    """ Calculate if dest_path has room to hold the specified elements.
+
+    Determine whether or not dest_path will exceed the specified usage ratio
+    (used:total space) after adding the specified elements.
+
+    Parameters
+    ----------
+    elements : sqlalchemy.orm.query.Query
+        The elements to be added to the destination directory
+
+    src_path : str
+        The path in which the elements are currently stored
+
+    dest_path : str
+        The path to which the elements would be moved.
+
+    ratio : double
+        The allowed ratio of used:total space.
+
+    Returns
+    -------
+    True if adding "elements" to "dest_path" does not increase the ratio past its
+      threshold.  Else false.
+
+    """
     if elements:
         addcount = 0
         size = 0
@@ -111,6 +135,17 @@ class QueueProcess():
 
 
     def get_logger(self, log_level):
+        """ Instantiate and return a logger based on process inforamtion.
+        Parameters
+        ----------
+        log_level : str
+            The string descriptor of the log level (critical, error, warning, info, debug)
+
+        Returns
+        -------
+        logger : logging.Logger
+            The parameterized logger
+        """
         logger = logging.getLogger(f"{self.process_name}_Queueing.{self.archive}")
         level = logging.getLevelName(log_level)
         logger.setLevel(level)
@@ -123,10 +158,12 @@ class QueueProcess():
 
 
     def get_matching_files(self):
+        """ Get the files matching the process characteristics.  Overridden by child classes."""
         raise NotImplementedError()
 
 
-    def enqueue(self, element):
+    def enqueue(self, element, copy):
+        """ Add an element to the queue.  Overridden by child classes."""
         raise NotImplementedError()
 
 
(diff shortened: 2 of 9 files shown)


pds_pipelines/queueing.py
```diff
@@ -0,0 +1,236 @@
...
+            size += getsize(fname)
+
+        size_free = disk_usage(dest_path).free
+        if size >= (disk_usage_ratio * size_free ):
+            return False
+    return True
+
+
+class QueueProcess():
+    def __init__(self, process_name, archive, volume=None, search=None, log_level='INFO'):
```
dpmayerUSGS, line 82: @AustinSanders Could we also add queue namespace as a parameter? We currently allow overriding the default namespace on the commandline in `service_job_manager.py`.

pds_pipelines/queueing.py
```diff
@@ -0,0 +1,337 @@
...
+        else:
+            results = session.query(Files).filter(Files.archiveid == self.archive_id,
+                                               Files.upc_required == 't')
+        if self.search:
+            qf = '%' + self.search + '%'
+            results = results.filter(Files.filename.like(qf))
+        session.close()
+        return results
+
+    def enqueue(self, element, copy):
```
AustinSanders, line 296: One thing that I'm not particularly happy with is the enqueue function.  Specifically, it seem to break the single responsibility principle in that it is responsible for both copying and queueing.  I wasn't able to find an elegant way of splitting up the functionality since the the filename that we're queueing comes from the copy function rather than the query object.  I don't think this would be too hard to fix, but I'm blanking.

pds_pipelines/queueing.py
```diff
@@ -0,0 +1,337 @@
...
+        copy : boolean
+            Copy files to the workspace if true.
+
+        Returns
+        -------
+        None
+        """
+        path = self.archive_info[self.archive]['path']
+        fname = path+element.filename
+        if copy:
```
AustinSanders, line 267: I don't know of any scenario in which we'd want to copy the files for DI processing, but the copy flag was necessary in order to make this fit the interface.  I think this is the natural consequence of deriving an interface from existing components rather than starting with a clean interface design upon which components are built.  I don't think this is _too_ big of an issue, but it's worth pointing out.

pds_pipelines/queueing.py
```diff
@@ -0,0 +1,337 @@
...
+
+        size_free = disk_usage(dest_path).free
+        if size >= (disk_usage_ratio * size_free ):
+            return False
+    return True
+
+
+class QueueProcess():
+    def __init__(self, process_name, archive, volume=None, search=None, log_level='INFO', namespace=None):
+        self.process_name = process_name
```
AustinSanders, line 112: I didn't put too much though into which of these should be member variables and which should be passed around the object.  Since this is an object that represents a process, I kind of just stuffed everything process-related in here.  Might be worth trimming down?

pds_pipelines/queueing.py
```diff
@@ -0,0 +1,337 @@
...
+        self.logger.info("%s queue: %s", process_name, self.process_queue.id_name)
+
+        try:
+            pds_session_maker, _ = db_connect(pds_db)
+            self.logger.info('Database Connection Success')
+        except Exception as e:
+            self.logger.error('Database Connection Error\n\n%s', e)
+            exit()
+
+        self.session_maker = pds_session_maker
```
AustinSanders, line 134: I could use a second opinion on this.  I decided to let the object manage its own sessions (rather than creating a session / sessionmaker and passing it to the object) because, no matter what, the queueing process requires a database connection.  Since queues are essentially pulling from the database and putting into redis queues, it made sense to me to allow the object to manage its own database / redis connections.

pds_pipelines/queueing.py
dpmayerUSGS: DI is purely about checking that the checksum in the database matches the checksum of the file *in the exact location* recorded in the database, so we would never need to copy the file as part of running DI.
That being said, I don't think there's an issue with the solution here.

dpmayerUSGS:
@AustinSanders I think `ingest_queueing.py` needs to be updated (and tested) as well.

pds_pipelines/queueing.py
AustinSanders: I resolved this with **kwargs.  There's still a flexible interface for enqueue, but it allows additional keyword arguments as needed.

pds_pipelines/queueing.py
```diff
@@ -0,0 +1,390 @@
...
+        """
+        logger = logging.getLogger(f"{self.process_name}_Queueing.{self.archive}")
+        level = logging.getLevelName(log_level)
+        logger.setLevel(level)
+        logFileHandle = logging.FileHandler(pds_log + 'Process.log')
+        formatter = logging.Formatter(
+            '%(asctime)s - %(name)s - %(levelname)s, %(message)s')
+        logFileHandle.setFormatter(formatter)
+        logger.addHandler(logFileHandle)
+        return logger
```
acpaquette, line 159: Could be changed to a has attribute model, something like:
```
if not (hasattr(self, logger)):
    ~get the logger~
    self.logger = logger
return self.logger
```

pds_pipelines/queueing.py
```diff
@@ -0,0 +1,390 @@
...
+        self.archive_info = json.load(open(pds_info, 'r'))
+        try:
+            self.archive_id = self.get_archive_id()
+        except KeyError:
+            self.logger.error("Archive %s not found in %s", archive, pds_info)
+            raise
+        self.volume = volume
+        self.search = search
+        self.namespace = namespace
+        self.process_queue = RedisQueue(f"{process_name}_ReadyQueue", namespace)
```
acpaquette, line 125: I don't think process queue is necessarily wrong here but it's a bit of a misnomer, maybe `ready_queue` instead?

pds_pipelines/queueing.py
```diff
@@ -0,0 +1,390 @@
...
+                        cast(Files.di_date, Date) is None))
+        else:
+            results = session.query(Files).filter(Files.archiveid == self.archive_id).filter(
+                or_(cast(Files.di_date, Date) < testing_date,
+                    cast(Files.di_date, Date) is None))
+        session.close()
+
+        return results
+
+    def enqueue(self, element, **kwargs):
```
acpaquette, line 256: kwargs isn't being used here. Is this just a signature thing or do the kwargs need to be pushed to the queue?

pds_pipelines/queueing.py
```diff
@@ -0,0 +1,390 @@
...
+        path = self.archive_info[self.archive]['path']
+        fname = path+element.filename
+        if copy:
+            fname = copy_files(path+fname, archive_base, workarea)
+        self.process_queue.QueueAdd((fname, self.archive))
+
+class IngestQueueProcess(QueueProcess):
+    def __init__(self, *args, link_only=False, **kwargs):
+        super().__init__(*args, **kwargs)
+        # Default to empty string if falsy value specified
```
acpaquette, line 347: falsy to false

acpaquette (commented):
Overall looks good! A couple questions and potential changes, dig the polymorphism to reduce redundant code!

pds_pipelines/queueing.py
acpaquette: I think this is fine. These problems tend to happen a fair amount when using an object oriented approach but I don't think it's out of the objects scope to query the DB since it needs it to build the set of items to send to the queue

pds_pipelines/queueing.py
acpaquette: I don't know if these objects necessarily represent a process, rather they build sets that represent processes. It more or less needs the process_name, archive, we could have setters for the rest but they could also remain as they are. It's more of a personal preference and really depends on the person programming it. I think this is great for a first pass and we can change things if we want to down the line

pds_pipelines/queueing.py
acpaquette: Was this fixed/addressed? It looks like there are some leftover lines of code that could be changed. The actual copy looks like it could be taken care of in the run function on the base object?

pds_pipelines/queueing.py
AustinSanders: This is just a signature thing in order to allow the enqueue(element, copy=copy) in the run loop.  It's kind of ugly, but I wasn't sure how else to handle scenarios in which _some_ queue types require a copy flag.

pds_pipelines/queueing.py
AustinSanders: As much as I hate it, "falsy" is a thing, and I feel that it's more applicable here. [reference](https://stackoverflow.com/questions/39983695/what-is-truthy-and-falsy-how-is-it-different-from-true-and-false)

pds_pipelines/available_modules.py
```diff
@@ -10,7 +10,7 @@
                  f'Failed with the following {e}.')
 
 try:
-    import gdal
+    from osgeo import gdal
```
acpaquette, line 13: I swear I have fixed this like two or three times now 😢 I would also update the ogr import

acpaquette (approved):
One small thing to fix but other than that this looks solid!

pds_pipelines/queueing.py
```diff
@@ -0,0 +1,359 @@
...
+        if size >= (disk_usage_ratio * size_free ):
+            return False
+    return True
+
+
+class QueueProcess():
+    def __init__(self, process_name, archive, volume=None, search=None, log_level='INFO', namespace=None):
+        self.process_name = process_name
+        self.archive = archive
+        self.logger = self.get_logger(log_level)
```
acpaquette, line 115: Technically this doesn't have to be set like this any more but it can also stay this way

pds_pipelines/available_modules.py
dpmayerUSGS: @AustinSanders IMO this is the only thing standing in the way of merging at this point

dpmayerUSGS (changes requested):
Once you update the gdal import in available_modules, this PR will be good to merge.

dpmayerUSGS (approved):

dpmayerUSGS:
Closes #469

Issue #497, opened 2020-09-13
write_view_jsons.py: Add more queries for min/max statistics and histograms of fields from search_terms table


dpmayerUSGS:
### Description ###
[write_view_jsons.py ](https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/46004032aeb91498c55a34b91a0211577072dd7f/pds_pipelines/write_view_jsons.py#L1) is a successful proof-of-concept for a way to query the UPC database and write various statistics to JSON, which can in turn be consumed by public facing services (such as PILOT) or used internally to provide summary insights into the contents of the UPC. 

In order for this script to evolve from a proof-of-concept to a production-ready tool, several queries should be added that will extract the following information:

#### Stats Per Target ####
- [ ] Total number of images 

#### Stats from `search_terms` Table, Per-Target, Per-Instrument ####
- [ ] The number of *mapped* images
  - has footprint in `isisfootprint` field of search_terms table
- [ ] The number of *unmapped* images
  -  isisfootprint is NULL
- [ ] Minimum and maximum values of the following fields:
  - [ ] `starttime` ([already implemented?](https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/46004032aeb91498c55a34b91a0211577072dd7f/pds_pipelines/write_view_jsons.py#L10)) 
  - [ ] `solarlongitude`
  - [ ] `meangroundresolution`
  - [ ] `minimumphase`
  - [ ] `maximumphase`
  - [ ] `minimumincidence`
  - [ ] `maximumincidence`
  - [ ] `minimumemission`
  - [ ] `maximumemission` 

#### Histogram Summaries from `search_terms` Table, Per-Target, Per-Instrument ####
It appears that histogram summaries are already implemented in the script for most of the fields in the search terms table. A histogram summary of `starttime` should be added, binned by Earth month over the total range of start times.

PR #529, opened 2021-01-22
added data query for json files

robotprogrammer22:
Added queries mentioned in #497.


dpmayerUSGS:
This should stay as a draft until we find a solution for handling histogram summary queries.

dpmayerUSGS (approved):

PR #531, opened 2021-01-26
UPC bugfix, small updates to PDSinfo.json and cleaning up recipe files

### LINKED ISSUE #532: upc_process.py: source URL being built incorrectly
## Description ##
https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/76588e1cf9b1a614256155b226d9411e918408d9/pds_pipelines/upc_process.py#L397-L401

As currently written, the code does not generate a correct URL. For example, if
```
inputfile = "/foo/bar/workarea/Mars_Odyssey/THEMIS/path/to/file.QUB"
workarea = "/foo/bar/workarea"
archive_base = "/pds_san/PDS_Archive"
archive_path = archive_base + "/Mars_Odyssey/THEMIS"
web_base = "https://pdsimage.wr.usgs.gov/Missions/"
```
Then the above code yields
```
orig_file = "/pds_san/PDS_Archive/Mars_Odyssey/THEMIS/Mars_Odyssey/THEMIS/path/to/file.QUB"
edr_source = "https://pdsimage.wr.usgs.gov/Missions/Mars_Odyssey/THEMIS/Mars_Odyssey/THEMIS/path/to/file.QUB"
```
which is incorrect (notice "Mars_Odyssey/THEMIS" appears twice in the URL).

## Expected Result ##
What we expect is `edr_source = "https://pdsimage.wr.usgs.gov/Missions/Mars_Odyssey/THEMIS/path/to/file.QUB"`

## Proposed Solution ##
The 4 lines of code quoted above should be replaced with a single line:
```
edr_source = inputfile.replace(workarea, web_base)
```
`orig_file` is not used anywhere else in the script, and none of the information from `PDSinfo.json` is used either.
This means that `upc_process.py` doesn't actually need to import and read `PDSinfo.json` at all, so the import statement can be removed from the top of the script as well.

### LINKED ISSUE #536: upc_process: Misnamed keyword prevents writing footprint to database
The `GisFootprint` keyword is no longer being found correctly during UPC processing, resulting in `isisfootprint` in the `search_terms` table always being NULL even if a footprint exists in the `caminfo` output.

I'm not sure if we switched to forcing keywords to be all lowercase in caminfo output, or if this is a change in behavior of the `pvl` library, but `GisFootprint` needs to be changed to `gisfootprint` on the following line:
https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/76588e1cf9b1a614256155b226d9411e918408d9/pds_pipelines/upc_process.py#L238

dpmayerUSGS:
Code changes to upc_process.py
Fixes #532 
Fixes #536 

Plus some housekeeping:
- Updated archiveid for THEMIS SDPs in PDSinfo.json
- Updated POW recipes for Viking Orbiter EDRs
- Updated UPC/reduced recipes for THEMIS IR/VIS and MGS MOC
- Removed unused "projected" section from recipe files
- Deleted (currently) unused recipe files
- Deleted an errant .swp file


USGSsakins (approved):
The changes look good.

PR #539, opened 2021-03-06
Fix get_single_band_cube() to grab correct band, logger format update, recipe updates

### LINKED ISSUE #537: get_single_band_cube() should search index+1 of band, not band number itself
https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/43f6c1b4c2a9f430ab26cffaee52e101be4a168b/pds_pipelines/available_modules.py#L108
## Description
`get_singleband_cube()` is meant to take a list of integers referring to band numbers for a given instrument and then run `cubeatt` to output an ISIS cube containing *only* data from the first matching band in the list. The integers in the input list are better understood as names for bands because for some instruments, such as THEMIS IR, not every image contains all possible bands, and the order in which the bands appear in the input image does not necessarily match the numerical order of the band names. 

For example, the lists of THEMIS IR band numbers in preferred order is `[9,8,7,6,5,4,3,2,1,10]`. A THEMIS IR image may only contain data from Bands 9, 8, and 7 (in that order). As currently written `get_singleband_cube()` won't find a a 9th band, but it will find a 3rd band, which actually is THEMIS IR Band 7, and return a cube containing Band 7 data, when we would prefer it to return Band 9 (the first band in the example input cube).

## Solution
The line at the top of the issue should be changed to look for the *index*(+1) of the list of band names from the input cube (`bands_in_cube`):
`isis.cubeatt(from_=cube + '+' + str(bands_in_cube.index(band) + 1), to=out_cube)`

### LINKED ISSUE #538: derived_process.py: Make logger format similar to upc_process.py
`derived_process.py` executes a process that logically follows from UPC processing. It would be helpful if the format of the logger output from `derived_process.py` matched that of `upc_process.py`

This could be done by replacing the follow lines
https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/43f6c1b4c2a9f430ab26cffaee52e101be4a168b/pds_pipelines/derived_process.py#L51-L57

with something more like analogous code from `upc_process.py`:
```
    try:
        slurm_job_id = os.environ['SLURM_ARRAY_JOB_ID']
        slurm_array_id = os.environ['SLURM_ARRAY_TASK_ID']
    except:
        slurm_job_id = ''
        slurm_array_id = ''


    inputfile = ''
    context = {'job_id': slurm_job_id, 'array_id':slurm_array_id, 'inputfile': inputfile}
    logger = logging.getLogger('Derived_Process')
    level = logging.getLevelName(log_level)
    logger.setLevel(level)
    log_file_handle = logging.FileHandler(pds_log + 'Process.log')
    formatter = logging.Formatter(
        '%(asctime)s - %(job_id)s - %(array_id)s - %(inputfile)s - %(name)s - %(levelname)s, %(message)s')
    log_file_handle.setFormatter(formatter)
    logger.addHandler(log_file_handle)
    logger = logging.LoggerAdapter(logger, context)
```

dpmayerUSGS:
Fix get_single_band_cube() bug and modify logger format in derived_process.py to match upc_process.py

Fixes #537 
Fixes #538 

Non-code changes:
Updated UPC and derived recipes for Galileo SSI, Mariner 10, MESSENGER MDIS, MGS MOC, THEMIS IR, and Viking Orbiter

### DIFF
--- a/pds_pipelines/available_modules.py
+++ b/pds_pipelines/available_modules.py
@@ -105,7 +105,7 @@ def get_single_band_cube(cube,out_cube,band_list,keyname):
 
     for band in band_list:
         if band in bands_in_cube:
-            isis.cubeatt(from_=cube + '+' + str(band), to=out_cube)
+            isis.cubeatt(from_=cube + '+' + str(bands_in_cube.index(band) + 1), to=out_cube)
             break
         else:
             continue
--- a/pds_pipelines/derived_process.py
+++ b/pds_pipelines/derived_process.py
@@ -8,6 +8,8 @@ import errno
 from ast import literal_eval
 from json import JSONDecoder
 
+from sqlalchemy import exc
+
 from pds_pipelines.redis_queue import RedisQueue
 from pds_pipelines.redis_lock import RedisLock
 from pds_pipelines.db import db_connect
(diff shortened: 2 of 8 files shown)


USGSsakins (approved):
This looks good, it seems like a few different things mixed together, do we want to separate these kinds of changes into separate PRs in the future?

dpmayerUSGS:
@USGSsakins Sure, I can split these up in the future. I should be better about opening issues for recipe changes as well.

PR #544, opened 2021-03-22
Fixes detached label derived processing

acpaquette:
Fixes a small query issue when processing a derived product from the upc database.


dpmayerUSGS (approved):
Looks good. +1 for making single line change vs. 6 line solution I suggested.

PR #545, opened 2021-03-22
Address inability to connect to the upc database

### LINKED ISSUE #534: upc_process.py (and derived_process.py?): Not catching db connection errors correctly
## Description ##
In the event of an unexpected loss of connection to a database, `upc_process.py` is supposed catch an exception, requeue the item to be processed, and exit gracefully:
https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/76588e1cf9b1a614256155b226d9411e918408d9/pds_pipelines/upc_process.py#L477-L480

However, when a loss of connection occurs, the script is crashing. Relevant section from error output:
```
psycopg2.OperationalError: server closed the connection unexpectedly
        This probably means the server terminated abnormally
        before or while processing the request.


The above exception was the direct cause of the following exception:

  File "/scratch/pds_services/workarea/ingest_backlog/PDS-Pipelines/pds_pipelines/upc_process.py", line 432, in main
    target_qobj = Targets.create(session, targetname=target_name,
  File "/home/pds2010mgr/anaconda3/envs/PDS-backlog-mgmt/lib/python3.9/site-packages/pds_pipelines/utils.py", line 205, in wrapper
    res = f(*args, **kwargs)
  File "/home/pds2010mgr/anaconda3/envs/PDS-backlog-mgmt/lib/python3.9/site-packages/pds_pipelines/models/upc_models.py", line 131, in create
    target_qobj = session.query(Targets).filter(
...
  File "/scratch/pds_services/workarea/ingest_backlog/PDS-Pipelines/pds_pipelines/upc_process.py", line 503, in <module>
    sys.exit(main(parse_args()))
  File "/scratch/pds_services/workarea/ingest_backlog/PDS-Pipelines/pds_pipelines/upc_process.py", line 474, in main
    except sqlalchemy.exc.SQLAlchemyError as e:
NameError: name 'sqlalchemy' is not defined
```

## Possible Solution ##
Is this as simple as changing the import statement at the top of the affected script to import the `exc` module from `sqlalchemy` and then changing the exception from `sqlalchemy.exc.SQLAlchemyError ...` to `exc.SQLAlchemyError ...` ?

## Additional Context ##
`derived_process.py` contains the same exception catch as `upc_process.py`. I haven't encountered this issue in `derived_process.py`, but I think that's just luck so far. A solution implemented for `upc_process.py` should be mimicked in `derived_process.py`.
I don't know how to reproduce this issue or simulate database connection errors.

### LINKED ISSUE #540: upc_process.py: Add "else" condition if database session not established
## Description
https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/088ef31a8cf7de68c2858d0f3d8c557276d8a4b0/pds_pipelines/upc_process.py#L425

`upc_process.py` only attempts to write information to the UPC database `if` it has been able to establish sessions to the UPC and DI databases, but there is no `else` action currently defined. Consequently, processing of a file will "fail silently" if the script has *not* been able to establish the database sessions.

## Proposed Solution
Add an `else` condition that prints a descriptive error message to the log and requeues the input file.

acpaquette:
Adds more informative error handling if a database could not be connected to. 

Also removes any use of the pds_session_maker, not sure why this was present in this processing script anymore.

Closes #540 
Closes #534

### DIFF
--- a/pds_pipelines/upc_process.py
+++ b/pds_pipelines/upc_process.py
@@ -15,6 +15,7 @@ from glob import glob
 import pvl
 import json
 from sqlalchemy import exc, and_
+from sqlalchemy.exc import SQLAlchemyError
 from pds_pipelines.available_modules import *
 from osgeo import ogr
 
@@ -29,7 +30,7 @@ from pds_pipelines.pvl_utils import load_pvl, find_keyword
 from pds_pipelines.db import db_connect
 from pds_pipelines.models import pds_models
 from pds_pipelines.models.upc_models import SearchTerms, Targets, Instruments, DataFiles, JsonKeywords, BaseMixin
-from pds_pipelines.config import pds_log, workarea, keyword_def, pds_db, upc_db, lock_obj, upc_error_queue, web_base, archive_base, recipe_base
+from pds_pipelines.config import pds_log, workarea, keyword_def, upc_db, lock_obj, upc_error_queue, web_base, archive_base, recipe_base
 from pds_pipelines.utils import generate_processes, process, add_process_db, get_isis_id
 
 
@@ -347,7 +348,6 @@ def parse_args():
 
 def main(user_args):
     upc_session_maker, upc_engine = db_connect(upc_db)
-    pds_session_maker, pds_engine = db_connect(pds_db)
 
     persist = user_args.persist
     log_level = user_args.log_level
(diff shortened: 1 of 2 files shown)


pds_pipelines/upc_process.py
```diff
@@ -15,6 +15,7 @@
 import pvl
 import json
 from sqlalchemy import exc, and_
+from sqlalchemy.exc import SQLAlchemyError
```
dpmayerUSGS, line 18: @acpaquette Could you update the corresponding import statement (and the later try/else block) in `derived_process.py` as well? Issue #534 affects both `upc_process.py` and `derived_process.py`

dpmayerUSGS (changes requested):

pds_pipelines/upc_process.py
acpaquette: I think this is being handled correctly in `derived_process.py`. The script imports `exc` and then uses `exc.SQLAlchemyError`

dpmayerUSGS (approved):
Looks good.

Issue #543, opened 2021-03-22
ingest_queueing.py: Fails with undefined variable when user passes invalid archive name


dpmayerUSGS:
`ingest_queueing.py` is supposed to return a list of valid archive names (based on contents of `PDSinfo.json` if the user passes an invalid archive name. This looks like an unanticipated consequence of some of the recent changes to queueing.

```
python ingest_queueing.py -a foobar

Traceback (most recent call last):
  File "/home/pds2010mgr/anaconda3/envs/PDS-Services/lib/python3.8/site-packages/pds_pipelines/queueing.py", line 222, in get_archive_id
    archive_id = self.archive_info[self.archive]['archiveid']
KeyError: 'lamp'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "ingest_queueing.py", line 46, in <module>
    sys.exit(main(parse_args()))
  File "ingest_queueing.py", line 40, in main
    process = IngestQueueProcess('Ingest', archive, volume, search, log_level, namespace)
  File "/home/pds2010mgr/anaconda3/envs/PDS-Services/lib/python3.8/site-packages/pds_pipelines/queueing.py", line 318, in __init__
    super().__init__(*args, **kwargs)
  File "/home/pds2010mgr/anaconda3/envs/PDS-Services/lib/python3.8/site-packages/pds_pipelines/queueing.py", line 118, in __init__
    self.archive_id = self.get_archive_id()
  File "/home/pds2010mgr/anaconda3/envs/PDS-Services/lib/python3.8/site-packages/pds_pipelines/queueing.py", line 224, in get_archive_id
    print("\nArchive '{}' not found in {}\n".format(archive, info))
NameError: name 'archive' is not defined

```

dpmayerUSGS:
Closed by #546

PR #546, opened 2021-03-23
Archive Error Message Fix

acpaquette:
Fixes archive error message, so that the available archives are displayed if an incorrect archive is entered.


dpmayerUSGS (approved):
Looks good.

dpmayerUSGS:
Fixes #543

PR #547, opened 2021-03-23
Database Gremlin Squashing #1

acpaquette:
Adds a context based session manager to mitigate branching session handling.

Changes automatic database creation/structure check to a manual function that needs to be called for new or munged databases.


dpmayerUSGS (approved):
I successfully queued and processed several hundred MGS MOC images through both upc_process and derived_process using this code.

PR #550, opened 2021-03-24
Queueing KeyError Removal

acpaquette:
Prevents KeyError from being displayed to the user of the queueing scripts


dpmayerUSGS (approved):
Looks like a perfectly reasonable solution to me.

PR #555, opened 2021-04-26
Updated UPC/Derived recipes for Mariner 10, MARCI, Voyager

### LINKED ISSUE #554: Update Voyager UPC recipe
The UPC and reduced recipes in `recipe/voyager.json` need to be updated to use the jinja template format. The recipes currently use the old format and therefore will cause `upc_process.py` and `derived_process.py` to fail when run.

dpmayerUSGS:
Fixes #554

### DIFF
--- a/recipe/mariner_10.json
+++ b/recipe/mariner_10.json
@@ -4,7 +4,7 @@
     "upc": {
         "recipe": {
             "isis.mar102isis": {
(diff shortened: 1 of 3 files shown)


USGSsakins (commented):
Looks good, the variable names are more self documenting, should be easier to support.

PR #563, opened 2021-08-16
Added check for `catlab` output in upc_process, updated LROC UPC recipe…

### LINKED ISSUE #562: upc_process.py: Speed up `pvl.load()`
https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/f2b083a0d149dafd64a5a081a78d94f2409a5cc8/pds_pipelines/upc_process.py#L428

## Description ##
The `pvl.load()` function can take a long time (minutes to hours) to load the PDS label from large files with embedded labels, such as LROC NAC EDRs.  `upc_process.py` only needs to extract a few items from the label, so it's wasteful to have to wait for `pvl.load()` to scan through a potentially large image file to only load the relatively small label.

## Desired Behavior ##
The call to `pvl.load()` on the line above should only take as long as required to load the label as it would if the label were detached.

## Proposed Solution ##
Allow UPC recipes to include a call to the ISIS `catlab` program to dump PDS labels to a standalone PVL file, and then check for the existence of such a file within `upc_process.py`. If present, load the standalone PVL file, else load the original input file as the script currently does.

dpmayerUSGS:
… to include `catlab` step

Fixes #562

### DIFF
--- a/pds_pipelines/upc_process.py
+++ b/pds_pipelines/upc_process.py
@@ -417,15 +417,24 @@ def main(user_args):
         no_extension_inputfile = os.path.splitext(inputfile)[0]
         cam_info_file = no_extension_inputfile + '_caminfo.pvl'
         footprint_file = no_extension_inputfile + '_footprint.json'
+        catlab_output = no_extension_inputfile + '_catlab.pvl'
 
         processes = generate_processes(inputfile,
                                        recipe_string, logger,
                                        no_extension_inputfile=no_extension_inputfile,
+                                       catlab_output=catlab_output,
                                        cam_info_file=cam_info_file,
                                        footprint_file=footprint_file)
         failing_command, _ = process(processes, workarea, logger)
 
-        pds_label = pvl.load(inputfile)
+        # Some datasets with attached PDS labels cause PVL to hang,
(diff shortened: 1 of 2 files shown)


thareUSGS:
Looks good. I assume for our purposes here there is no reason to calibrate the images (thus the step was removed).

dpmayerUSGS:
@thareUSGS No calibration step was removed; I don't know what you're referring to. UPC recipes never involve running a calibration step. The LROC recipe was consolidated to have `caminfo` generate the footprint (because we have to run `caminfo` anyway), rather than doing a separate call to `footprintinit`.

thareUSGS:
I was just noticing that it looks like in the files changed page, lronaccal was removed from recipe lro_lroc_edr.json? https://github.com/USGS-Astrogeology/PDS-Pipelines/pull/563/files#diff-0782d0f34232b0ae5cb90a28b272fa7cb3c9da38ed8218700d24c6fc1951b2e2

dpmayerUSGS:
Oh that's down in the browse/thumbnail section, not UPC. I thought that was pulled out a while ago. I wonder if I didn't sync the repo properly before submitting the PR? The recipe looks correct after the merge, so no harm done.

PR #584, opened 2021-09-14
 POW: Correctly apply user-supplied crop/trim extents

### LINKED ISSUE #581: `service_job_manager.py`: Map extents not being applied to POW jobs
There appear to be 2 related bugs within `service_job_manager.py` that prevent crops from being applied correctly to POW jobs if the user specifies a crop extent. The result is that POW produces output images with an extent determined by the "DEFAULTRANGE=MINIMIZE" option to `cam2map` rather than being cropped or trimmed to the extent entered by the user.

The last `elif` in the following block should be split into separate cases for rangetype = smart and rangetype=fill. In both cases `recipeOBJ['isis.cam2map']['defaultrange']` to be set to "MAP" rather than "CAMERA." Additionally, for the rangetype = fill case, the `cam2map` TRIM option should be set to YES (`recipeOBJ['isis.cam2map']['trim'] = 'YES'`) 
https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/8bfa67612b1041fe441792662cc3ccf985de8d50/pds_pipelines/service_job_manager.py#L626-L628

The following block adds min/max lat/long values to the map template file for MAP2 files _only_. This is incorrect; the outer `if` statement should be removed and the child `if` statements should be de-dented so that that are replied regardless of whether the job is a POW or MAP2 job.
https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/8bfa67612b1041fe441792662cc3ccf985de8d50/pds_pipelines/service_job_manager.py#L967-L975

### LINKED ISSUE #582: `di_process.py` Incorrect number in log statement
https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/8bfa67612b1041fe441792662cc3ccf985de8d50/pds_pipelines/di_process.py#L95-L97

DI runs `session.commit()` after the index > 250, but the statement that is printed to the log still refers to the previous threshold value of 50. The number in the log statement should be changed to the correct value of 250.

### LINKED ISSUE #583: `PDSinfo.json`: Remove `upc_reqs` from `mo_themis_sdp`
https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/8bfa67612b1041fe441792662cc3ccf985de8d50/pds_pipelines/PDSinfo.json#L189

The THEMIS SDP archive is complicated because it contains multiple product types, some of which are UPC-eligible and some which are not, so it's not possible to specify a single set of `upc_reqs` to flag all of the UPC-eligible files on Ingest and _avoid_ flagging files that are _not_ UPC-eligible. 

Until/unless we develop a more sophisticated logic for identifying UPC-eligible products on Ingest, a workaround would be to remove the `upc_reqs` keyword for `mo_themis_sdp` and *manually* update the DI database after each Ingest to set the UPC flag, then rely on the existing `mo_themis_ir_edr` and `mo_themis_vis_edr` entries in `PDSinfo.json` to map the different product types to their corresponding UPC/POW recipes.

dpmayerUSGS:
Fixes #581
Also made 2 non-functional changes:
- Correct typo in a print statement (Fixes #582)
- Removed `upc_reqs` from `mo_themis_sdp` (Fixes #583)


thareUSGS:
Good catch as this is a little used setting for POW (but good to have).

PR #591, opened 2021-10-05
CI Workflow

acpaquette:
Adds workflow for continuous integration on PR into master, thanks autocnet!


jlaura (approved):

dpmayerUSGS (approved):
Thanks for cleaning up that if/else block in `upc_process.py` too

PR #600, opened 2021-10-07
Added Namespace Arguments to Process Scripts

### LINKED ISSUE #564: {ingest/upc/derived/di}_process scripts: Add ability to specify queue namespace on command line
## Description
This is issue is a natural follow-on to a feature that was added to several of the queuing scripts a while back. The queueing scripts for Ingest, UPC, Derived, and DI allow the user to specify a custom namespace for the Redis queues on the command line. For example: `ingest_queuing.py -a lro_lroc_edr -v LROLRC_0001 -n mynamespace` will add items to a queue named "mynamespace:Ingest_ReadyQueue" instead of teh default "queue:Ingest_ReadyQueue." However, the corresponding process scripts are still hard-coded to process items from queues under the default namespace defined in config.py. This is a bug because even though a user can add items to a queue under a custom namespace, there's no way to actually process these items.

## Proposed Solution
Add a command line option, `--namespace` to `{ingest,upc,derived_di}_process.py`, similar to the option that already exists for the corresponding queuing scripts. If the user passes this option and a custom namespace, propagate this to `RedisQueue()` within the processing script, otherwise use the default namespace.

amystamile-usgs:
Closes #564


pds_pipelines/derived_process.py
```diff
@@ -68,8 +80,13 @@ def main():
...
 
-    RQ_derived = RedisQueue('Derived_ReadyQueue')
-    RQ_error = RedisQueue(upc_error_queue)
+    if namespace:
+        RQ_derived = RedisQueue('Derived_ReadyQueue', namespace)
+        RQ_error = RedisQueue(upc_error_queue, namespace)
+    else:
+        RQ_derived = RedisQueue('Derived_ReadyQueue')
+        RQ_error = RedisQueue(upc_error_queue)
+
```
acpaquette, line 89: I think we can just set namespace to `None` by default and avoid the duplicate calls without the namespace. Checkout the `__init__` function for `RedisQueue` in `redis_queue.py`. It handles None, by using the default_namespace. This idea applies to all of the queue creation updates in this PR

acpaquette (commented):
A couple code suggestions to limit duplicate calls

AustinSanders (approved):

PR #601, opened 2021-10-08
Split processing/updating into two scripts, combine upc/derived processing

### LINKED ISSUE #574: Queue Files for Derived Processing from `upc_process.py` by Default, but Make Optional
## Description
The UPC workflow involves the following steps:

1. `upc_queueing.py -a <archive name> -v <volume name>` to copy eligible files to the work area and add their paths to UPC_ReadyQueue
2. `hpc_job_manager.py -p upc` to submit array of `upc_process.py` tasks to SLURM
3. `derived_queueing.py -a <archive name> -v <volume name>` to copy *the same files from Step 1* to the work area and add their paths to Derived_ReadyQueue
4. `hpc_job_manager.py -p derived` to submit array of `derived_process.py` tasks to SLURM

(Derived processing depends on the existence of a record for a given file in the UPC database, so it is only valid to run `derived_process.py` *after* `upc_process.py` has been run on a given file.)

### Pain Points
- In practice, derived queueing/processing for a given archive and volume is performed immediately following completion of UPC processing and the input files for UPC processing remain in the work area until they're manually deleted (this is a feature, not a bug). This means that `derived_queueing.py` is copying files into the work area that are already there. This is unnecessary and wastes time (for some archives, > 12 hours for a typical archive volume). 

- `derived_queueing.py` runs the same query against the DI database that `upc_queueing.py` runs in order to populate the Redis queue associated with the process. This doesn't waste as much time as (re)copying files, but it shouldn't be necessary if we've just run the query for the purpose of populating the UPC_ReadyQueue.

- Within some datasets, the wall time required to run a single image can be highly variable--minutes for some images, several days for others--although the ones that take days usually represent <1% of the files in a UPC run. (The long processing time is typically due to `caminfo` taking a long time on images with weird viewing geometries.) This "long tail" of processing time means that a few images may still be running through UPC processing while tens of thousands of images are done. Derived processing on the volume gets held up waiting for just a few images because if a user runs `derived_queueing.py` and starts derived processing as described above while the UPC process is still running, it will queue the UPC "stragglers" that don't yet have records in the UPC database, causing those images to fail derived processing.

### Wants
As member of the PDS team who has to run an end-to-end UPC workflow, I want to only have to copy files once from the archive area into the work area. I also want to have the option to have `upc_process.py` add a file to Derived_ReadyQueue as it completes the initial UPC processing so that I can potentially start derived processing while a small number of "stragglers" are still running through UPC processing without fear that entries in Derived_ReadyQueue don't yet have UPC database entries. Put another way, I would like to eliminate Step 2 from the procedure above for most scenarios, but retaining the ability to follow the procedure above in scenarios where files *do* need to be re-copied to the work area for derived processing.

## Proposed Solution
I think this can be solved by adding a command line option and associated logic to `upc_process.py`. Specifically, add code that pushes the item being worked on from UPC_WorkQueue (see #568) onto Derived_ReadyQueue and then removes it from UPC_WorkQueue . This should become the default behavior for `upc_process.py`, but there should also be a command line option (i.e. `--no-derived`) that suppresses pushing an item onto Derived_ReadyQueue.

### LINKED ISSUE #588: Split the image processing and database update portions of `{upc,derived}_process.py` into separate scripts
This issue is a component of implementing #580 . See that issue for a high-level overview.

### Option A: 2 scripts-->4 scripts
Based on how `{upc,derived}_process.py` are currently written, it would be easy to split these 2 scripts into 4,  i.e. `{upc,derived}_process.py` and `{upc,derived}_update.py`. This is the lowest-effort solution and will meet the requirements/desires of #580. 

### Option B: Consolidate Image Processing into Single Script, Database Updates into Single Script, but with Granular Options
If Option A is implemented the resulting image processing scripts would look nearly identical, save for which Redis queue they key off of (UPC_ReadyQueue vs. Derived_ReadyQueue) and which key they read from a recipe file ("upc" vs. "derived"). It may make more sense to consolidate the image processing into a single script and use commandline arguments to select whether the script should read/run the UPC or Derived recipe, or both (i.e. `--upc`, `--derived`, `--all`). Similarly, the database update actions could be consolidated into a single script than can create an initial UPC record, or update an existing record with the browse/thumbnail URLs, or both. 

### Option C: Eliminate Concept of "Derived" as a Separate Process Altogether
In practice, the Derived recipes consist of running `*2isis` for a given dataset and then running `gdal.Translate()` a couple times to make downsampled JPEGs. The database update portion consists of inserting 2 key:value pairs (containing URLs to the browse and thumbnail images) into the "json_keywords" table of the UPC database. This is so simple that perhaps it makes more sense to eliminate the concept of "Derived" as a separate process within the broader UPC workflow and instead generate the browse/thumbnail images as part of the existing upc recipes, and include inserting the browse/thumbnail URLs into the json_keywords table as part of the database update portion. Under this option, `upc_process.py` would still be split to leave the image processing in place and spin the database update code off into `upc_update.py`, where the later would consolidate the database update functionality of the existing `{upc,derived}_process.py` scripts. 

## Gotchas
If Option B is implemented, is Derived_ReadyQueue still necessary? Probably not. It's possible that a user will start image processing on one dataset and want *both* the upc and derived recipes run, but at the same time another user will start image processing on a different dataset requesting *only* the derived recipe be run. An easy way to enforce separation between the 2 jobs in this scenario would be to have each job consume a UPC_ReadyQueue under a *different namespace.*  The scripts should have an option to specify a non-default queue namespace on the commandline (cf. #564).

If Option C is implemented, all of the existing recipe files would need to be updated to fold the GDAL translate steps from the "derived" section into the "upc" section. Any of the Redis queues named like "Derived_*" would be eliminated as obsolete because image processing would be accomplished by consuming "UPC_ReadyQueue" and database updates by consuming (yet-to-be-implemented) "UPC_UpdateQueue."

AustinSanders:
Closes #574 Closes #588

Combines upc/derived processing and splits up the processing / update portions of those scripts.  This eliminates the need for derived queueing.  Also includes a change that allows the user to select which processing is desired via --no-upc and --no-derived flags.  Default functionality is that both types of processing are enabled.

This should not be merged until I can resolve the merge conflicts that I expect to come from #600. I'll mark this as a draft until then.

I had to pull from an existing PR (#593) in order to get the local database up and running.  Those commits should disappear from this PR when it is merged.


pds_pipelines/upc_update.py
```diff
@@ -0,0 +1,556 @@
...
+                    target_qobj = Targets.create(session, targetname=target_name,
+                                                 displayname=target_name.title(),
+                                                 system=target_name)
+                    target_id = target_qobj.targetid
+
+
+
+                # Some datasets with attached PDS labels cause PVL to hang,
+                #  so recipe includes call to dump label using `catlab`
+                # If present, use the catlab output as pds_label instead of inputfile
```
dpmayerUSGS, line 487: This comment block belongs immediately before Line 469 (the if/else block where `pds_label` gets defined)

environment.yml
```diff
@@ -7,7 +7,7 @@ dependencies:
     - psycopg2
     - lxml
     - pytz
-    - sqlalchemy=1.3
+    - sqlalchemy
```
dpmayerUSGS, line 10: What are the effects of unpinning sqlalchemy (other than upgrading to version 1.4.x)? 
Might be better to leave it pinned until we can test thoroughly.

acpaquette: This was largely an issue where our environments were pulling 1.1.14 for whatever reason so a long while ago I pinned it to 1.3. If our tests comeback okay then this is probably okay to un-pin. We have tests that hit the databases, so we should know if this causes an issue. I would like to test it before on the production database just to make sure that it works with whatever version of postgis we are running there.

pds_pipelines/upc_process.py
```diff
@@ -376,25 +54,19 @@ def main(user_args):
...
-
     # Redis Queue Objects
     RQ_main = RedisQueue('UPC_ReadyQueue', namespace)
     RQ_error = RedisQueue(upc_error_queue, namespace)
     RQ_work = RedisQueue('UPC_WorkQueue', namespace)
+    RQ_update = RedisQueue('UPC_UpdateQueue', namespace)
 
     logger.info("UPC Processing Queue: %s", RQ_main.id_name)
 
+    RQ_error = RedisQueue(upc_error_queue)
```
amystamile-usgs, line 65: Looks like RQ_error was added twice here.

pds_pipelines/upc_process.py
```diff
@@ -1,359 +1,37 @@
...
-    return json_keywords_attributes
 
 def parse_args():
     parser = argparse.ArgumentParser(description='UPC Processing')
-    parser.add_argument('--persist', '-p', dest="persist",
-                        help="Keep intermediate .cub files.", action='store_true')
+    parser.add_argument('--no-upc', dest='proc', help='Do not perform UPC processing',
+                        action="store_false")
+    parser.add_argument('--no-derived', dest='derived', help='Do not perform derived processing',
+                        action="store_false")
```
amystamile-usgs, line 20: I see that you did `store_false` for these "no flags". For consistency, I am okay with the idea of having all these flags have the `store_false` action. I think if all the negative flags are consistent in the code base, that we would not have problems with readability. If everyone is okay with that idea, I can update my existing PR to match these.

dpmayerUSGS (approved):

PR #604, opened 2021-10-14
Removed automatic root config

### LINKED ISSUE #467: Config "root" uses installed root location, not package root location
After installing the package with pip, I used ```python thumbnail_queueing.py -a mro_ctx -v mrox_2854 -s J12_049471_1414_XN_38S165W```.  This gave the error:

FileNotFoundError: [Errno 2] No such file or directory: '/home/pds2010mgr/anaconda3/envs/austin_pds/lib/python3.8/site-packages/logs/Process.log'

It seems that the config uses the installed script's location, not the current / working directory.

AustinSanders:
closes #467

### DIFF
--- a/pds_pipelines/config.py
+++ b/pds_pipelines/config.py
@@ -47,8 +47,7 @@ link_dest = '/pds_san/PDS_Archive_Links/'
 
 # The project root.  The 'main' folder containing source, log, and output subdirectories.
 #  e.g. '/home/some_user/pds_pipelines' will lead to logs stored in '/home/some_user/pds_pipelines/logs'
-root = Path(__file__).parent.parent
-
+root = '/home/some_user/pds_pipelines'
 
 ##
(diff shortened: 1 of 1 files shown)


pds_pipelines/config.py
```diff
@@ -47,8 +46,7 @@
 
 # The project root.  The 'main' folder containing source, log, and output subdirectories.
 #  e.g. '/home/some_user/pds_pipelines' will lead to logs stored in '/home/some_user/pds_pipelines/logs'
-root = Path(__file__).parent.parent
```
amystamile-usgs, line 50: I wonder if this is worth keeping as is for testing purposes but putting a comment that this should be updated to your local project root path.

amystamile-usgs (approved):

dpmayerUSGS (approved):

PR #607, opened 2021-10-15
Refactored upc_process and upc_queueing for threadsafe updates

AustinSanders:
Refactored upc_process and upc_queueing for threadsafe updating.

Previously, upc_process produced and enqueued two update jobs, one for the upc update and one for derived products.  This was problematic because the derived product update depended on a upcid being in the database.

This commit refactors upc_process so that it only produces and enqueues a single job.  If the job is of type 'upc,' then upc_update will add the upc information and attempt to input derived information.  If no derived products are found, then it skips that portion.  If the job is of type 'derived,' then upc_update will search for a matching upcid and update the record with the derived information.

### DIFF
--- a/pds_pipelines/upc_update.py
+++ b/pds_pipelines/upc_update.py
@@ -520,7 +520,8 @@ def main(user_args):
                 with session_scope(upc_session_maker) as session:
                     JsonKeywords.create(session, **json_keywords_attributes)
 
-            # Always attempt derived processing
+            # Derived Processing:
+
             # If we don't have a upcid, get the matching ID from the database
             if not 'upc_id' in locals():
                 with session_scope(upc_session_maker) as session:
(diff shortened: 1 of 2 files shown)


pds_pipelines/upc_update.py
```diff
@@ -520,29 +520,42 @@ def main(user_args):
                 with session_scope(upc_session_maker) as session:
                     JsonKeywords.create(session, **json_keywords_attributes)
 
-            #derived
-            elif update_type.lower() == 'derived':
+            # Derived Processing:
+
+            # If we don't have a upcid, get the matching ID from the database
+            if not 'upc_id' in locals():
```
dpmayerUSGS: Instead of checking for upc_id in the local symbol table, consider initializing `upc_id = None` _at the top of the while loop_ and then doing `if not upc_id: ...`
I think it would make this line easier to understand.

This will also help ensure that a upc_id value doesn't get carried over from one iteration of the loop to the next.

PR #616, opened 2022-01-20
Optimized band_summary_query

### LINKED ISSUE #615: UPC Database Schema: Add Band Summary Table
https://github.com/USGS-Astrogeology/PDS-Pipelines/blob/de2df8de59d5a0c8017c61874f0a304254a8e1bc/pds_pipelines/write_view_jsons.py#L29-L37

## Description
The band summary query in `write_view_jsons.py` times out for certain instruments (including Clementine) after ~2 hours. The query is run on a per-mission basis inside a FOR loop, so whichever instrument it fails on causes the script to exit with an error and prevents queries from being run on the subsequent instruments.

## Desired Behavior
The `write_view_jsons.py` script should successfully output band summary stats for all instruments in the UPC database.

## Proposed Solution
I think the volume of data is just too big to query efficiently using the current approach. One alternative solution would be to implement a _band summary table_ in the UPC database, as exists in the "old" UPC. However, fully resolving this issue will require not only updating the database schema, but making some code changes in user-facing scripts as well:

- [ ] Update schema to include bands_summary table
- [ ] Add function to `upc_update.py` that inserts new filter name and center wavelengths as needed (similar to existing functions to insert instruments and targets)
- [ ] Update `write_view_jsons.py` to query new band_summary table directly to output JSON files on per-mission basis

### bands_summary Columns
The legacy bands_summary table is indexed on `instrumentid` from the instruments table and then has "filter" as a varchar field to hold the filter or band name, and "center" as a double precision field to hold the center wavelength value.

dpmayerUSGS:
Updated the band summary query in `write_view_jsons.py` to perform joins on pre-filtered data so that the queries can run successfully as the database grows large. 

Fixes #615 via a simpler approach to the one proposed in the issue.


acpaquette (approved):
Good catch closing the sessions and the proper join requirements work as expected!

